PR #3, opened 2021-03-12
Upgrade flight-job to have standalone feature parity with the api

WilliamMcCumstie:
The models and templates have been ported from `flight-job-api` so the CLI can be used as a standalone utility.

This adds various new `script` and `job` commands which implement the feature parity. The command line has also received work to add a `--json` flag.


foo.out
```diff
@@ -0,0 +1 @@
+hello world
```
benarmston, line 1: Why do we have `foo.error` and `foo.out` files committed to the repo?

lib/flight_job/command.rb
```diff
@@ -25,72 +25,118 @@
...
+        interactive: (opts.ascii || opts.pretty || $stdout.tty? ? true : nil),
+        json: (opts.json ? true : nil)
+      }
     end
 
     def load_template(name_or_id)
+      template = Template.new(id: name_or_id)
+      if template.exists?
+        raise InternalError, <<~ERROR.chomp unless template.valid?(:verbose)
+          Can not load the following template as it is invalid: #{template.id}
```
benarmston, line 76: s/Can not/Cannot/

lib/flight_job/command.rb
```diff
@@ -25,72 +25,118 @@
...
           ERROR
         end
-        templates[index]
-
-      # Handles an exact match
-      elsif match = templates.find { |t| t.name == name_or_id }
-        match
+        templates[index].tap do |template|
+          raise InternalError, <<~ERROR unless template.valid?(:verbose)
+            Can not load the following template as it is invalid: #{template.id}
```
benarmston, line 94: *Cannot

lib/flight_job/help_formatter.rb
```diff
@@ -0,0 +1,225 @@
...
+# changes are made available under the EPL-2.0 mentioned above. The original
+# work is available under the following licenses.
+#
+# The following license applies to changes made subsequent to the point
+# this version was forked from the original upstream version available
+# at <https://github.com/commander-rb/commander>.
+#
+# The MIT License (MIT)
+#
+# Copyright (c) 2017-present Alces Flight Ltd <licensing@alces-flight.com>
```
benarmston, line 37: Why do we have two licences for Alces Flight Ltd?

libexec/slurm/sbatch.sh
```diff
@@ -0,0 +1,13 @@
...
+# NOTE:
+#
+# Please make installation specific modifications within this file.
+# The last line printed to STDOUT for a successful deployment must be the
+# response from sbatch:
+#
+# $ sbatch my-script
+# > Submitted batch job XX
+#-------------------------------------------------------------------------------
+sbatch "$1"
```
benarmston, line 13: What's the purpose of this file?

benarmston (commented):
A few comments for you to look at @WilliamMcCumstie .

lib/flight_job/help_formatter.rb
WilliamMcCumstie: Because I pulled the `help_formatter.rb` from `commander-openflighthpc` which is licensed under MIT to be consistent with the original `commander` gem.

We *maybe* able to change the alces component of the license to `EPL`? However, I haven't put much thought into the legality of doing so.

Eventually this code will likely be backported to `commander`, so I didn't feel it was worth dealing with licensing details.

libexec/slurm/sbatch.sh
WilliamMcCumstie: This is a legacy implementation from `flight-job-script-api`.

The `submit.sh` contains a bunch of boiler plate code concerning the JSON response the tool is expecting. This means the `submit.sh` system managed and does not support edits.

So a compromise to this was to create the `sbatch.sh` which is packaged as as config file and can be modified with installation specific details.

foo.out
WilliamMcCumstie: 95efe852ac5976e47000b00ede3f254b0b92a091

lib/flight_job/command.rb
WilliamMcCumstie: 95efe852ac5976e47000b00ede3f254b0b92a091

lib/flight_job/command.rb
WilliamMcCumstie: 95efe852ac5976e47000b00ede3f254b0b92a091

lib/flight_job/render_context.rb
```diff
@@ -0,0 +1,87 @@
...
+        @table = Hash.new(&b)
+      else
+        @table = {}
+      end
+      @table.define_singleton_method(:key?) { |_| true }
+      opts.each do |k, v|
+        @table[k.to_sym] = v
+      end
+    end
+  end
```
benarmston, line 86: What's this class for?  In particular, why is `key?` being defined to always return `true`?

lib/flight_job/version.rb
```diff
@@ -25,5 +25,5 @@
 # https://github.com/openflighthpc/flight-job
 #==============================================================================
 module FlightJob
-  VERSION = '1.1.0'
+  VERSION = '1.2.0'
```
benarmston, line 28: This needs updating.

benarmston (changes requested):
I have a few more things for you to look at here please.


* The cron job is still not working.  The monitor script is ran by cron, but it doesn't change the status of the job.  If I run the monitor script manually it works correctly.

* Selecting multiple values form the multi select was non-obvious.  Could something be done to add some help text explaining how to use the multi select?

* We should remove `deprecated` from the help text.  I don't see why `copy-template` is deprecated.  Either its useful and should be kept or it isn't useful and should be removed.  I think its probably useful.

* What's the purpose of the `run-monitor` command.  Both the cron job and the API run a monitor script directly instead of running that command.

---

Not one to consider for this PR as its related to the template content rather than the functionality, but I'll add it here as this is where I'll remember to look for it.

* Setting the working directory to `/home/%u` doesn't work.  The `%u` is not expanded to the user's name.

lib/flight_job/render_context.rb
WilliamMcCumstie: The `DefaultOpenStruct` provides some syntactic sugar to the rendering. In particular it provides handle for missing questions. For example:

```
# If a standard OpenStruct is used
questions.missing               # => nil
questions.missing.answer  # => NoMethodError

# When using the DefaultOpenStruct
question.missing               # => AnswerDecorator
question.missing.answer  # => nil
```

The `OpenStruct` object does not support the defaults mechanism provided by `Hash`. It also duplicates the input `Hash`, which prevents defaults being support via the original hash.

Also because there are infinitely many "missing keys", the standard `defaults_hash.merge(answers)` mechanism does not work.

So instead the `DefaultsOpenStruct` bridges the gap between the two. It is an `OpenStruct` that supports the defaults mechanism provided by `Hash`.

In regards to the `key?` method, because it is an "OpenStruct" with defaults, it responds to all keys. Thus the `key?` method should always be `true`.

benarmston: Thanks for the response.  Could you add a brief comment to this affect to the class.

I wonder if we'll eventually want to differentiate between a question existing or not.  If we do, we might want to change the behaviour of `key?`.  But let's leave that for future us to worry about.

WilliamMcCumstie:
The following changes have been made in regards to issues identified below

1. As discussed IRL, the `crontab` issue is an integration issue with `flight-slurm`. As this is an installation specific issue it is best dealt with in the builder
2. The prompt colours have been modified to make multi-selects a bit more obvious. Try it out and let me know what you think,
3. Deprecated warning has been removed,
4. The `run-monitor` command is the CLI entry point and is useful if the user wants to force an update. The `run-monitor` script has been moved to `libexec` to denote it is internal. The purpose of the script is to make integrating with the `crontab` easier and not to be user facing. The CLI is still responsible for the heavy lifting.

> I have a few more things for you to look at here please.
> 
> * The cron job is still not working.  The monitor script is ran by cron, but it doesn't change the status of the job.  If I run the monitor script manually it works correctly.
> * Selecting multiple values form the multi select was non-obvious.  Could something be done to add some help text explaining how to use the multi select?
> * We should remove `deprecated` from the help text.  I don't see why `copy-template` is deprecated.  Either its useful and should be kept or it isn't useful and should be removed.  I think its probably useful.
> * What's the purpose of the `run-monitor` command.  Both the cron job and the API run a monitor script directly instead of running that command.
> 
> Not one to consider for this PR as its related to the template content rather than the functionality, but I'll add it here as this is where I'll remember to look for it.
> 
> * Setting the working directory to `/home/%u` doesn't work.  The `%u` is not expanded to the user's name.

lib/flight_job/version.rb
WilliamMcCumstie: https://github.com/openflighthpc/flight-job/pull/3/commits/9e7aa3bc0fb68fb8ddfce896d8b10f03c04cc051

lib/flight_job/render_context.rb
WilliamMcCumstie: For future reference,

Changing the behaviour of `key?` whilst still inheriting off `OpenStruct` is probably a bad idea. I believe that `OpenStruct` uses `key?` internally so I'm not sure what side effects it will introduce.

In general this is a quick and dirty way to implement the rendering features we want. There is probably scope for replacing it with a dedicated object at some point in the future.

WilliamMcCumstie: 19dde4487630fad7b6bec3d862690badd39e724b

libexec/slurm/monitor.sh
```diff
@@ -0,0 +1,114 @@
...
+if echo "$start_time" | grep -P "$format" ; then
+  start_time="$start_time$(date +%:z)"
+else
+  start_time=""
+fi
+if echo "$end_time" | grep -P "$format" ; then
+  end_time="$end_time$(date +%:z)"
+else
+  end_time=""
+fi
```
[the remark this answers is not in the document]
WilliamMcCumstie: https://github.com/openflighthpc/flight-job/pull/3/commits/735da9d46ec52c7298b7dfe165ff1cce2eb73eb1

libexec/slurm/sbatch.sh
benarmston: What installation specific details do you imagine would be needed?  Perhaps this would be better renamed to `sbatch_wrapper.sh` to communicate its purpose.

benarmston:
Thanks for the changes @WilliamMcCumstie .

The multiselect output is much easier to understand.

I've added (what should be a final) question about the `sbatch.sh` wrapper.  Once that is dealt with I think this is good to go.

libexec/slurm/sbatch.sh
WilliamMcCumstie: Renamed here: https://github.com/openflighthpc/flight-job/pull/3/commits/da2842f1aea2a1d5fd84f351f2865fc28cedc472

There could be various flags that a sysadmin might want to set without worrying about updating every `template`/`script`. The most likely candidates are `--cluster`, `--priority`, and possible `--time`.

Or they may want to preform some bespoke action depending on who the user is. The wrapper provides a useful integration point getting caught up in the `submit.sh` JSON handling.

In general a more flexible "submit" mechanism may be preferred, however the requirements have not been established for this.

PR #4, opened 2021-03-23
Add a command to edit scripts

WilliamMcCumstie:
A new `edit-script` command has been added to edit scripts. The script is opened in the user's `$EDITOR` or `$VISUAL` editor before defaulting to `vi`.

When using `vi`/`vim`/`nvim`, the editor will open at the start of the user's workload section. This is done via the use of a magic comment: `#====>><<====` (or various similar).

These comments have been embedded in the templates and will become available in new rendered templates.

Based on #3, please merge it first


benarmston:
I like the idea of opening the editor at the start of the workload.  We can better determine where the workload starts from the line

```
#  >>>> YOUR WORKLOAD
```

That puts the environment module loading visible in the editor and that is likely something that the user will want to change.

WilliamMcCumstie:
> I like the idea of opening the editor at the start of the workload. We can better determine where the workload starts from the line
> 
> ```
> #  >>>> YOUR WORKLOAD
> ```
> 
> That puts the environment module loading visible in the editor and that is likely something that the user will want to change.

In principle I have no issue moving the "anchor point" higher up, however I have a small concern using `YOUR WORKLOAD` as the delimiting characters. Mainly it provides semantic information in which the template author may want to change. For example, a more highly customised `template` may want to describe the workload (e.g. `Model 2.1 OpenFoam Workload`).

Alternatively a purely syntax delimiting pattern is ideal, possible:

```
# >>>> foobar <<<<
```

This way the template author has a bit more control over how the label the template sections. It also allows them to move the opening point or disable it entirely.

lib/flight_job/cli.rb
```diff
@@ -100,6 +100,16 @@ def self.create_command(name, args_str = '')
       c.slop.bool '--stdin', 'Provide the answers via STDIN as JSON'
     end
 
+    create_command 'edit-script', 'SCRIPT_ID' do |c|
+      c.summary = 'Open the script in your editor'
+      c.description = <<~DESC.chomp
+        Open and update your script within the systems editor.
+
+        The editor can be selected using the $EDITOR or $VISUAL
+        environment variable. 'vi' will be used if neither is set.
```
benarmston, line 109: How about:

```
Edit your script.

Open the script in the editor given by `$VISUAL`, `$EDITOR` or `vi`.

Changes you make will affect any future jobs submitted from this script, but will not affect jobs already submitted.
```

lib/flight_job/commands/edit_script.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    def script
+      @script ||= load_script(args.first)
+    end
+
+    # Determine which line to open the script on
+    def start_line
+      File.open(script.script_path) do |file|
+        _, idx = file.each_line.each_with_index.find do |line, _|
+          />{4,}.*<{4,}/.match?(line)
```
benarmston, line 60: This pattern match doesn't work for any scripts I created in prior versions of flight-job.  Matching against `/^# *>{4,}.*WORKLOAD` should be robust against most changes to templates and still work with the existing scripts.

lib/flight_job/commands/edit_script.rb
```diff
@@ -0,0 +1,76 @@
...
+  class Commands::EditScript < Command
+    def run
+      # Ensure the script exists up front
+      script
+
+      # Opens vim/emacs/nano commands at the start line, else use the regular editor
+      # NOTE: gedit needs x-forwarding over ssh, but it also supports the +line
+      cmd = if ['vi', 'vim', 'nvim', 'emacs', 'nano', 'gedit'].include?(editor)
+              if start_line
+                "#{editor} +#{start_line}"
```
benarmston, line 40: Oh dear! You've got me thinking of how awesome this could be:

For `vim` and `nvim` we could add the following to the command:

```
-c 'let b:scrollofforig=&scrolloff' -c 'set scrolloff=0' -c 'normal! kztj' -c 'let &scrolloff=b:scrollofforig'
```

Without those additions the editor opens as this:


image


With those editions it opens as:

image


A little thing, but it really does focus the user's attention on their workload.

benarmston: I'm sure there are options for emacs too, but I don't know them.  If you do, I'd be pleased to have them included, if you don't I'm happy to ship without too.

benarmston (changes requested):
This looks mostly good to me.  Just an issue about supporting existing scripts that needs to be looked at, and an option to make the vim experience even more awesome.

lib/flight_job/commands/edit_script.rb
WilliamMcCumstie: :laughing: I barely supported `emacs` in the first place!

benarmston:
I've added a commit to remove half of the funky vim open commands I suggested in https://github.com/openflighthpc/flight-job/pull/4#discussion_r601644112.  I've kept the half that actually did something.  The part removed was an attempt to work around a setting that didn't quite work.

With that removed, this looks good to me.

PR #6, opened 2021-03-29
Allow the script name to be set

WilliamMcCumstie:
This PR splits the old random `id` into two new components:

* A user friendly `public_id` which the user can optionally set/change. It defaults to the `template_name` using `wget` style, and
* A static `internal_id` (UUID) which never changes. This is used for referential integrity.


benarmston:
@WilliamMcCumstie can you rebase this on `master` please.

WilliamMcCumstie:
Rebased, I think

WilliamMcCumstie:
Okay after much rebasing, this branch has been integrated with the current state of master.

Also I wasn't happy with how the `public_id` and `internal_id` was working. I ran into a couple of problems:
* `rename-script` was too complex and risked duplicating the internal_id files,
* Determining the `metadata_path` was hard due to it being dependent on `public_id` which can change,
* The `internal_id` lookups where an O(S) whilst being preformed more often the `public_id` lookups. This meant caching was required to make `job` loads a O(S+J) operation.

Instead the `internal_id` and `script_id` have been switched around, otherwise the work broadly the same way. The notable differences are:

* The `scripts` are stored at: `./scripts/<internal-id>/{metdata.yaml,...}`
* The `internal_id_path` became the `public_id_path`: `./scripts/<internal-id>/public-<public-id>`
* The reservations files have been moved to a dedicated directory: `./scripts/reservations/public-<public-id>`

NOTE: Reservation files where moved so `scripts` are no longer made in a pseudo "allocated but non-existent" state. The reservation path still needs to match the script path's so it appears in the globs.

WilliamMcCumstie:
This is the wrong exit code:

```
COMMAND: /opt/flight/bin/flight job create-script simple biz-name --json --answers @-
USER: root
PID: 7252
STATUS: 3
STDIN:
{"notification_events":["Foo","ALL"],"id":"simple"}
STDOUT:

STDERR:
flight job: Failed to create 'biz-name' as it is already in used.
```

It should be `DuplicateError` and exit `7`.

benarmston:
This all seems very clever.  However, I think that it is too clever.  The PR seems to have strayed into reimplementing features that better belong to a RDBMS.  Notably, transactions and unique constraints.

Aside from the increased complexity, my biggest worry with this is the process killing in the `Script.reserve_public_id` method.

It looks to be OK as long as `flight job` is only ever ran on a single machine.  However, as `flight job` stores its data on a partition that is typically an NFS export, we could have a user run `flight job` from multiple machines sharing that data.  If that happened, we could end up killing a process that has nothing to do with `flight job`.  This could potentially result in a user losing data for a job that has been queued and/or running for several days.

This is a big enough concern that this PR can't be merged as it is.

Other issues that I have include:

* The `--internal-script-id` should be an internal abstraction, yet it leaks into the CLI.  The option isn't hidden, so users will see it and wonder what it is.  This could either lead to confusion or to experimentation and unintended consequences.

* The complexity in the `CreateScript` command has increased quite dramatically.

* Perhaps orthogonal to this PR, but there was already a lack of clarity with regard to a script's ID vs the `script_name` metadata item.  Replacing `id` with `public_id` and `internal_id` doesn't reduce this, instead it increases it.

---

The best route forward here is to implement much more limited functionality for now and later switch our data store to a RDBMS, allowing us to leverage its functionality to implement the more complicated features. 

I'm going to close this PR.  I'd like you to create a new PR implementing the following:

1. Optionally, allow for the script ID (aka name) to be provided when the script is created.  If one is not provided, the current mechanism of generating a short meaningless ID should be used.
2. Validation to ensure the given ID (aka name) is short.  Let's try with 16 chars and see how that works.
3. If an ID is provided, a check that it is not already used should be made.  No reservation system is required.
4. Once all of the answers, and notes have been collected, the script is created.  If a name has been provided and it is no longer unique, we exit with an informative error message without creating the script.

WilliamMcCumstie:
For reference:

> Aside from the increased complexity, my biggest worry with this is the process killing in the Script.reserve_public_id method.
>
> It looks to be OK as long as flight job is only ever ran on a single machine. However, as flight job stores its data on a partition that is typically an NFS export, we could have a user run flight job from multiple machines sharing that data. If that happened, we could end up killing a process that has nothing to do with flight job. This could potentially result in a user losing data for a job that has been queued and/or running for several days.

Assuming this comment is about the following line, the assessment isn't entirely correct:

https://github.com/openflighthpc/flight-job/blob/a068193d0574206515f4e72562b1d0f21360b2bc/lib/flight_job/models/script.rb#L92

It is essentially the same as running `kill -0 PID`, whilst looking nasty, it is actually reasonable benign. It does not sent any signals to the process, but instead preforms the existence/ permissions check.

With that said, I do agree it _may_ cause issues over NFS but not for your reasons stated. Instead the file locking will not work correctly as the two machines do not share a process list. Thus it will appear if the other machine's reservation has terminated.

This essentially breaks the reservation mechanism over NFS. So agree there is an issue, just for a different reason.

WilliamMcCumstie:
Closing so it can be re-implemented

PR #7, opened 2021-04-01
Tweak output rendering

WilliamMcCumstie:
The time fields have been standardised as `Created at`, `Started at`, `Ended At`.

---

The `StdOut Path` and `StdErr Path` are now combined together as `Output Path` if they are the same in the simplified human readable output.

This proved to be rather tricky as the rendering library (`OuputMode`) assumes that the `headers` are static. This assumption makes it easier to determine the padding of the headers.

In this particular case, the padding issue can be avoided as `StdOut Path` and `OutputPath` are the same length. This allows the internal template to switch out the header with an if-statement. However the padding will break it the lengths are different.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -48,7 +48,7 @@ GEM
       rexml
     method_source (1.0.0)
     minitest (5.14.4)
-    output_mode (1.4.0)
+    output_mode (1.5.0)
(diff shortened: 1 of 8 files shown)


benarmston:
I've changed the output handling slightly and added commits to deal with some different issues.  This looks good to merge now.

PR #8, opened 2021-04-12
Allow scripts to have friendly names

WilliamMcCumstie:
The `create script` command has been updated to allow them to have friendly IDs. It is not possible to rename existing scripts as this will break referential integrity between and `jobs` that are started from them.

### DIFF
--- a/lib/flight_job/models/script.rb
+++ b/lib/flight_job/models/script.rb
@@ -85,15 +85,15 @@ module FlightJob
     end
 
     validate on: :render do
-      # Ensures the metadata does not exists
-      if File.exists? metadata_path
-        @errors.add(:metadata_path, 'already exists')
+      # Attempt to create the directory for the script
+      # NOTE: The metadata_path/script_path does not need checking as they live
+      #       in this directory.
+      begin
+        FileUtils.mkdir_p FlightJob.config.scripts_dir
+        FileUtils.mkdir File.expand_path(id, FlightJob.config.scripts_dir)
+      rescue Errno::EEXIST
+        @errors.add(:id, 'already exists')
         next
-      end
-
-      # Ensures the script does not exists
-      if File.exists? script_path
-        @errors.add(:script_path, 'already exists')
       end
 
       # Ensures the template is valid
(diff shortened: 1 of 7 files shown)


lib/flight_job/models/script.rb
```diff
@@ -59,9 +59,18 @@ def self.load_all
...
     end
 
-    attr_writer :id, :notes
+    attr_reader :id
+    attr_writer :notes
+
+    validates :id, presence: true, length: { maximum: FlightJob.config.max_id_length },
+              format: { with: /\A[a-zA-Z0-9_-]+\Z/,
+                        message: 'must be alphanumeric but may contain hyphen and underscore' }
+    validates :id, format: { with: /\A[[:alnum:]].*\Z/, message: 'must start with an alphanumeric character' }
```
benarmston, line 68: For consistency with the above, "must start with a letter or a number".

lib/flight_job/models/script.rb
```diff
@@ -222,6 +241,17 @@ def render
...
+      # NOTE: As this is done after validation, it may trigger a race condition
+      #       This could cause the command to fail, whilst the other succeeds
+      #       Subsequent commands will detect the ID has been taken
+      begin
+        FileUtils.mkdir_p FlightJob.config.scripts_dir
+        FileUtils.mkdir File.expand_path(id, FlightJob.config.scripts_dir)
+      rescue Errno::EEXIST
+        raise InternalError, "Unexpectedly failed to claim ID: #{id}"
+      end
+
```
benarmston, line 254: I don't think this works as it should.  If the id has already been taken the `render` method will raise an error due to `valid?(:render)` returning false.  The error message presented to the user is `flight job: Unexpectedly failed to render the script!` instead of `flight job: Unexpectedly failed to claim ID: #{id}`.

benarmston (changes requested):
This looks much better than the previous PR.  I've added a few comments inline for you to look at.

lib/flight_job/models/script.rb
WilliamMcCumstie: 65f796c08732a8fd277b7bc00162554fca0a0d2f

lib/flight_job/models/script.rb
```diff
@@ -59,9 +59,18 @@ def self.load_all
       end.reject(&:nil?)
     end
 
-    attr_writer :id, :notes
+    attr_reader :id
+    attr_writer :notes
+
+    validates :id, presence: true, length: { maximum: FlightJob.config.max_id_length },
+              format: { with: /\A[a-zA-Z0-9_-]+\Z/,
+                        message: 'must be alphanumeric but may contain hyphen and underscore' }
```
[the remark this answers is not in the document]
WilliamMcCumstie: 65f796c08732a8fd277b7bc00162554fca0a0d2f

benarmston:
LGTM.

PR #10, opened 2021-04-16
Allow templates to be sorted in a priority oder

WilliamMcCumstie:
A new optional `priority` field has been added to `templates`. This is used to sort the `templates` into a custom order according to the following rules:

1. `templates` with a `priority` are sorted in ascending order,
2. `templates` with the same `priority` are sorted alphanumerically by their `id`, and
3. `templates` without a `priority` are sorted to the end (obeying condition 2).

---

The default templates have all been assigned either `priority` 1000 or 2000:
* `1000`: `simple` and `simple-array`, and
* `2000`: `mpi-nodes`, `mpi-slots`, and `smp`.

The `(0)-999` and `1001-1999` ranges have been left empty to allow custom templates to be sorted before and between the default blocks, respectively. Technically negative priorities are supported, however there use is discouraged to avoid confusing the end users.

Otherwise the `1000` and `2000` priorities are completely arbitrary. Making them configurable would be tricky as the `priority` needs to be hardcode into the `metadata.yaml` files for the `templates`.

By default, templates without a priority appear at the end of the list. This _feels_ like the appropriate location for these un-prioritised `templates`, however I am open to alternative suggestions on how they can be handled. The `priority` metadata key needs to be optional, as introducing new required keys is a breaking change.

### DIFF
--- a/lib/flight_job/models/template.rb
+++ b/lib/flight_job/models/template.rb
@@ -48,7 +48,8 @@ module FlightJob
             "required" => ['text', 'value'],
             "properties" => {
               'text' => { "type" => "string" },
-              'value' => { "type" => "string" }
+              'value' => { "type" => "string" },
+              'priority' => { "type" => 'integer' }
             }
           }
         }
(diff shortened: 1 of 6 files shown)


benarmston:
I've adjusted the priorities and [updated the client](https://github.com/openflighthpc/flight-job-script-service/commit/17430679e11384fc302acf03b528aa836aa3c8aa) to respect them.

PR #11, opened 2021-04-21
Prompt for the name when creating the script

WilliamMcCumstie:
The `id`/"name" of the script is now prompted for in an interactive session.

The `name` prompt is displayed after the summary as it may depend on the given answers/notes. It has been integrated into the `prompt_again` mechanism so it can be change.

In addition, various tweaks have been made to the format of the `summary` and the defaults to the prompts. This should hopefully improve the work flow.

### DIFF
--- a/bin/job
+++ b/bin/job
@@ -52,6 +52,14 @@ begin
   # Builds and runs the CLI
   require_relative '../lib/flight_job/cli'
 
+  # Kill attempt to kill all child processes on exit, this does two things:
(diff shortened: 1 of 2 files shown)


benarmston:
This works quite nicely.  Going forwards, we're likely going to either want to add support for editing the script itself as part of this loop or remove support for editing the notes.  Being able to edit one, but not the other doesn't quite seem right.  However, that is a future task.

PR #13, opened 2021-04-23
Map the slurm states to flight-job states

WilliamMcCumstie:
A mapping file of all slurm states has been added. It allows `scontrol` to return any state and map it to a known `flight-job` internal state. This prevents the proliferation of states for the time being, however some states (such as `TIMEOUT`) might need to be incorporated into `flight-job` proper.

A new concept of `RUNNING_STATES` has been added. These indicate the job has actually started and the `start_time` should be saved.

### DIFF
--- a/etc/flight-job.yaml
+++ b/etc/flight-job.yaml
@@ -66,6 +66,18 @@
 # monitor_script_path: libexec/slurm/monitor.sh
 
 # ==============================================================================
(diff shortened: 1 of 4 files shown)


benarmston:
I've updated the scheduler map file and added a commit to prevent the scheduler state from being lost.  After those this LGTM.

PR #14, opened 2021-04-26
Fix start_time for cancelled jobs and sort scripts/jobs

WilliamMcCumstie:
The list outputs are now sorted according to:

* `scripts`: By the `id`/`name` as this can be set by the end user, and
* `jobs`: By the `created_at` time

---

The `start_time` time is no longer set for jobs which are cancelled whilst in a pending state. This is handled by the `slurm/monitor.sh` script to prevent issues with polling intervals.

Jobs which were not allocated resources are "assumed" not to have started. This condition is detected in two ways:
* With `scontrol` when the `NodeList` is empty (technically `(null)`), and
* When `AllocTRES` field is empty in `sacct`.

### DIFF
--- a/libexec/slurm/monitor.sh
+++ b/libexec/slurm/monitor.sh
@@ -49,14 +49,14 @@ read -r -d '' template <<'TEMPLATE' || true
 TEMPLATE
 
 # Fetch the state of the job
-control=$(scontrol show job "$1" --oneline | head -n 1 | tr ' ' '\n' 2>&1)
+raw_control=$(scontrol show job "$1" --oneline 2>&1)
+exit_status="$?"
+control=$(echo "$raw_control" | head -n 1 | tr ' ' '\n')
 cat <<EOF
-Processed scontrol:
-$control
+scontrol:
+$raw_control
 EOF
 
-
-exit_status="$?"
 if [[ "$exit_status" -eq 0 ]]; then
   state=$( echo "$control" | grep '^JobState=' | cut -d= -f2)
   reason=$(echo "$control" | grep '^Reason='   | cut -d= -f2)
@@ -73,16 +73,16 @@ if [[ "$exit_status" -eq 0 ]]; then
   fi
 
   end_time=$(  echo "$control" | grep '^EndTime='   | cut -d= -f2)
-elif [[ "$control" == "slurm_load_jobs error: Invalid job id specified" ]]; then
+elif [[ "$raw_control" == "slurm_load_jobs error: Invalid job id specified" ]]; then
   # Fallback to sacct if scontrol does not recognise the ID
-  acct=$(sacct --noheader --parsable --jobs "$1" --format State,Reason,START,END,AllocTRES  | head -n1)
+  raw_acct=$(sacct --noheader --parsable --jobs "$1" --format State,Reason,START,END,AllocTRES)
+  exit_status="$?"
+  acct=$(echo "$raw_acct" | head -n1)
   cat <<EOF
 
-  First line of sacct:
-  $acct
+sacct:
+$raw_acct
 EOF
-
-  exit_status="$?"
 
   # Transition the job to "UNKNOWN" is sacct has no record of it
   if [[ "$exit_status" -eq 0 ]] && [ -z "$acct" ]; then
@@ -111,7 +111,6 @@ EOF
   fi
 else
   # Exit the monitor process if scontrol fails to prevent the job being updated
-  echo "$control" >&2
   exit "$exit_status"
 fi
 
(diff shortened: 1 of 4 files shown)


libexec/slurm/monitor.sh
```diff
@@ -49,21 +49,40 @@ read -r -d '' template <<'TEMPLATE' || true
...
 
 # Fetch the state of the job
-control=$(scontrol show job "$1" --oneline | head -n 1 | tr ' ' '\n' 2>&1)
+raw_control=$(scontrol show job "$1" --oneline 2>&1)
 exit_status="$?"
+control=$(echo "$raw_control" | head -n 1 | tr ' ' '\n')
+cat <<EOF
+scontrol:
+$control
+EOF
```
benarmston, line 58: Is this debugging code that should be removed?

libexec/slurm/monitor.sh
```diff
@@ -49,21 +49,40 @@ read -r -d '' template <<'TEMPLATE' || true
...
   # Fallback to sacct if scontrol does not recognise the ID
-  acct=$(sacct --noheader --parsable --jobs "$1" --format State,Reason,START,END  | head -n1)
+  raw_acct=$(sacct --noheader --parsable --jobs "$1" --format State,Reason,START,END,AllocTRES)
   exit_status="$?"
+  acct=$(echo "$raw_acct" | head -n1)
+  cat <<EOF
+
+sacct:
+$raw_acct
+EOF
```
benarmston, line 85: Is this debugging code that should be removed?

libexec/slurm/monitor.sh
```diff
@@ -73,11 +92,18 @@ elif [[ "$control" == "slurm_load_jobs error: Invalid job id specified" ]]; then
...
 
   # Extract the output from sacct
-  elif [[ "$exit_status" -eq 0 ]]; then
-    state=$(echo "$acct" | cut -d'|' -f1)
-    reason=$(echo "$acct" | cut -d'|' -f2)
+elif [[ "$exit_status" -eq 0 ]]; then
+  # sacct sometimes tacks info onto the end of the state
+  state=$(echo "$acct" | cut -d'|' -f1 | cut -d' ' -f1)
+  reason=$(echo "$acct" | cut -d'|' -f2)
+  if [ "$(echo "$acct" | cut -d'|' -f5)" ]; then
```
benarmston, line 99: It would be better to make the comparison to the empty string explicit here.  `if [ "$( ... )" != "" ]; then`.

benarmston (changes requested):
Could you take a look at the comments I've added @WilliamMcCumstie .

libexec/slurm/monitor.sh
WilliamMcCumstie: It is debugging code, but I decided to leave it in so it becomes part of the log file when `log_level` is set to `DEBUG`.

The JSON payload is limited to the last line, so it doesn't make a difference. However I can redirect it to STDERR as an alternative.

benarmston: Please redirect it to STDERR.

libexec/slurm/monitor.sh
benarmston: ping.

libexec/slurm/monitor.sh
benarmston: ping

libexec/slurm/monitor.sh
benarmston: ping

libexec/slurm/monitor.sh
WilliamMcCumstie: https://github.com/openflighthpc/flight-job/pull/14/commits/8dbac91b8f831737f4e38e773d163dac7a27bbc5

libexec/slurm/monitor.sh
WilliamMcCumstie: https://github.com/openflighthpc/flight-job/pull/14/commits/8dbac91b8f831737f4e38e773d163dac7a27bbc5

libexec/slurm/monitor.sh
WilliamMcCumstie: :facepalm:

WilliamMcCumstie: https://github.com/openflighthpc/flight-job/pull/14/commits/bd68dd00f25388794df831dc5bc8e97ca9087545

libexec/slurm/monitor.sh
WilliamMcCumstie: https://github.com/openflighthpc/flight-job/pull/14/commits/bd68dd00f25388794df831dc5bc8e97ca9087545

libexec/slurm/monitor.sh
WilliamMcCumstie: The idiomatic way to do this in `bash` is to use the `-n` flag:

https://github.com/openflighthpc/flight-job/pull/14/commits/34c9a3cffe21b9d5dd7790319e979f5f9a271433

benarmston: `-n` is good too.  As long as we're not relying on the reader to recall that `[ "" ]` is false.

benarmston:
LGTM now.

PR #16, opened 2021-04-27
Correctly terminate the process after interrupt

WilliamMcCumstie:
This stops the pager from breaking the terminal

### DIFF
--- a/bin/job
+++ b/bin/job
@@ -63,4 +63,5 @@ rescue Interrupt
   else
     $stderr.puts "\nWARNING: Cancelled by user"
   end
+  raise SignalException.new(2)
 end
--- a/bin/job
+++ b/bin/job
@@ -59,9 +59,9 @@ begin
   end
 rescue Interrupt
   if Kernel.const_defined?(:Paint)
-    $stderr.puts "\n#{Paint['WARNING', :underline, :yellow]}: Cancelled by user"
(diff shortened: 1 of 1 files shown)


bin/job
```diff
@@ -59,8 +59,9 @@ begin
   end
 rescue Interrupt
   if Kernel.const_defined?(:Paint)
-    $stderr.puts "\n#{Paint['WARNING', :underline, :yellow]}: Cancelled by user"
+    $stderr.print "\n#{Paint['WARNING', :underline, :yellow]}: Cancelled by user"
```
benarmston, line 62: Why the change from `puts` to `print`?

WilliamMcCumstie: > See for full details:
> https://trello.com/c/9sWZKxne/282-ttypager-breaks-the-terminal-on-interrupt#comment-60884aa45fc8de123ca1e421

tl;dr The `raise SignalException.new(2)` causes an additional line to be emitted. So switching to `print` accounts for it.

benarmston:
This is definitely better than it was before.  I'm getting behaviour where the first keypress after is ignored, but the terminal doesn't need resetting.  I'm happy to merge this once my question above is answered.

bin/job
WilliamMcCumstie:
```
$ irb
irb(main):001:0> raise SignalException.new(2)

$
```

PR #17, opened 2021-05-05
Track output files for the job 

WilliamMcCumstie:
The `submit.sh` script now reports back the `output_dir` where the job files are stored. The `output_dir` from the `submit.sh` script needs to match the `OUTPUT_DIR` within the rendered scripts.

For simplicity, this is done via convention. For `slurm` jobs, the `OUTPUT_DIR` must be:
`${WORKING_DIR}/${SLURM_JOB_NAME}-outputs/$SLURM_JOB_ID`.

It is "technically" possible to change the convention by re-configuring the application's `submit.sh` script. However this would require updating all the templates.

It maybe possible pass a custom the `OUTPUT_DIR` line when the scripts are rendered. This would allow the application to be easily reconfigured with a new convention.

---

Two new commands have been added: `ls-job-dir` and `view-job-file`.

The `ls-job-dir` wraps the `ls` utility and can pass down arbitrary flags. This does require the use of `--` delimiter.

The `view-job-file` opens the output files relative to the directory in the pager.

### DIFF
--- a/lib/flight_job/cli.rb
+++ b/lib/flight_job/cli.rb
@@ -183,15 +183,20 @@ module FlightJob
       c.summary = 'Display details about a submitted job'
     end
 
+    create_command 'view-job-file', 'JOB_ID FILENAME' do |c|
+      c.summary = "View a file within the job's output directory"
+    end
+
     create_command 'delete-job', 'JOB_ID' do |c|
       c.summary = 'Permanently remove a job'
     end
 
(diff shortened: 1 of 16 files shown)


benarmston:
I've added commits to fix the templates and rename the commands.  It turns out we can't set `WORKING_DIR` or `OUTPUT_DIR` amongst the slurm directives.

PR #18, opened 2021-05-05
Save the estimated start and end times

WilliamMcCumstie:
Because `scontrol` does not distinguish between "predicted" and "actual" start times, the state has to be interrogated to determine which one it is. This is done by `flight-job` ruby code as it has access to the state map.

The `submit.sh` script may always report a estimated start/end time and allow `flight-job` to determine how to handle it. The estimated start time can only updated in the `PENDING` state, after this point thus job has started and thus no more predictions are required. Similarly, the estimated finished time can only be updated in a `PENDING` or `RUNNING` state.

### DIFF
--- a/lib/flight_job/outputs/info_job.rb
+++ b/lib/flight_job/outputs/info_job.rb
@@ -117,10 +117,10 @@ module FlightJob
     #
     # Consider reordering on the next major version bump.
     register_attribute(header: 'Results Dir') { |j| j.results_dir }
-    register_attribute(verbose: true, header: 'Estimated Start') do |job|
+    register_attribute(verbose: true, header: 'Estimated start') do |job|
       job.format_estimated_start_time(true)
     end
-    register_attribute(verbose: true, header: 'Estimated Finish') do |job|
+    register_attribute(verbose: true, header: 'Estimated end') do |job|
       job.format_estimated_end_time(true)
     end
 
--- a/lib/flight_job/outputs/info_job.rb
+++ b/lib/flight_job/outputs/info_job.rb
@@ -34,7 +34,7 @@ module FlightJob
     TEMPLATE = <<~ERB
       <%
         verbose = output.context[:verbose]
-        main = output.callables.config_select(:section, :main)
+        main = output.callables.config_select(:section, :other)
         paths = main.select(&:paths?)
 
         # Determine if the STDOUT/STDERR paths should be combined or
@@ -66,17 +66,17 @@ module FlightJob
       <% end -%>
     ERB
 
-    register_attribute(section: :main, header: 'ID') { |j| j.id }
-    register_attribute(section: :main, header: 'Script ID') { |j| j.script_id }
-    register_attribute(section: :main, header: 'Scheduler ID') { |j| j.scheduler_id }
-    register_attribute(section: :main, header: 'State') { |j| j.state }
+    register_attribute(header: 'ID') { |j| j.id }
+    register_attribute(header: 'Script ID') { |j| j.script_id }
+    register_attribute(header: 'Scheduler ID') { |j| j.scheduler_id }
+    register_attribute(header: 'State') { |j| j.state }
 
     # Show a boolean in the "simplified" output, and the exit code in the verbose
-    register_attribute(section: :main, header: 'Submitted', verbose: false) { |j| j.submit_status == 0 }
+    register_attribute(header: 'Submitted', verbose: false) { |j| j.submit_status == 0 }
     # NOTE: There is a rendering issue of integers into the TSV output. Needs investigation
-    register_attribute(section: :main, header: 'Submit Status', verbose: true) { |j| j.submit_status.to_s }
+    register_attribute(header: 'Submit Status', verbose: true) { |j| j.submit_status.to_s }
 
-    register_attribute(section: :main, header: 'Submitted at') do |job, verbose:|
+    register_attribute(header: 'Submitted at') do |job, verbose:|
       if verbose
         job.created_at
       else
@@ -84,25 +84,11 @@ module FlightJob
       end
     end
 
-    # NOTE: These could be the predicted times instead of the actual, consider
-    # delineating the two
-    register_attribute(section: :main, header: 'Started at') do |job, verbose:|
-      if job.start_time.nil?
-        nil
-      elsif verbose
-        job.start_time
-      else
-        DateTime.rfc3339(job.start_time).strftime('%d/%m/%y %H:%M')
-      end
+    register_attribute(header: 'Started at') do |job, verbose:|
+      job.format_start_time(verbose)
     end
-    register_attribute(section: :main, header: 'Ended at') do |job, verbose:|
-      if job.end_time.nil?
-        nil
-      elsif verbose
-        job.end_time
-      else
-        DateTime.rfc3339(job.end_time).strftime('%d/%m/%y %H:%M')
-      end
+    register_attribute(header: 'Ended at') do |job, verbose:|
+      job.format_end_time(verbose)
     end
 
     # NOTE: In interactive shells, the STDOUT/STDERR are merged together if they
@@ -113,9 +99,9 @@ module FlightJob
     #       affecting the ability to pad the output.
     #
     # PS: The 'path' mode is a misnomer, it refers solely to the standard output/error paths
-    register_attribute(section: :main, modes: [:paths], header: 'Stdout Path') { |j| j.stdout_path }
-    register_attribute(section: :main, modes: [:paths], header: 'Stderr Path') { |j| j.stderr_path }
-    register_attribute(section: :main, interactive: true, modes: [:combined], header: 'Output Path') { |j| j.stdout_path }
+    register_attribute(modes: [:paths], header: 'Stdout Path') { |j| j.stdout_path }
+    register_attribute(modes: [:paths], header: 'Stderr Path') { |j| j.stderr_path }
+    register_attribute(interactive: true, modes: [:combined], header: 'Output Path') { |j| j.stdout_path }
 
     register_attribute(section: :submit, header: 'Submit Stdout') do |job|
       job.submit_stdout
--- a/lib/flight_job/outputs/info_job.rb
+++ b/lib/flight_job/outputs/info_job.rb
@@ -72,22 +72,22 @@ module FlightJob
     end
 
     start_header = ->(job, verbose:) do
-      job.start_time || verbose ? 'Started at' : 'Estimated Start'
+      job.actual_start_time || verbose ? 'Started at' : 'Estimated Start'
     end
     register_attribute(header: start_header) do |job, verbose:|
-      if job.start_time || verbose
-        job.format_start_time(verbose)
+      if job.actual_start_time || verbose
+        job.format_actual_start_time(verbose)
       else
         job.format_estimated_start_time(false)
       end
     end
 
     end_header = ->(job, verbose:) do
-      job.end_time || verbose ? 'Ended at' : 'Estimated Finish'
+      job.actual_end_time || verbose ? 'Ended at' : 'Estimated Finish'
     end
     register_attribute(header: end_header) do |job, verbose:|
-      if job.end_time || verbose
-        job.format_end_time(verbose)
+      if job.actual_end_time || verbose
+        job.format_actual_end_time(verbose)
       else
         job.format_estimated_end_time(false)
       end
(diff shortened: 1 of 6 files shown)


benarmston:
@WilliamMcCumstie can you rebase this on `master` please.  I attempted to do so, but I got bogged down in output mode section headers.

WilliamMcCumstie:
> @WilliamMcCumstie can you rebase this on `master` please. I attempted to do so, but I got bogged down in output mode section headers.

Rebased,

The `section` stuff provides control over heading padding, but I'm not particularity happy with how it works. I am in the process of phasing it out, which is why you ran into the conflicts.

libexec/slurm/monitor.sh
```diff
@@ -119,13 +128,17 @@ fi
...
+  estimated_start_time=""
+fi
 if [[ "$end_time" == "Unknown" ]] ; then
   end_time=""
 fi
 
 # Render and return the payload
 echo '{}' | jq  --arg state "$state" \
                 --arg reason "$reason" \
+                --arg estimated_start_time "$estimated_start_time" \
```
benarmston, line 141: Do we not need a similar line for the estimated end time?

lib/flight_job/outputs/info_job.rb
```diff
@@ -66,56 +53,55 @@ module Outputs::InfoJob
...
       end
     end
-    register_attribute(section: :main, header: 'Ended at') do |job, verbose:|
-      if job.end_time.nil?
-        nil
-      elsif verbose
-        job.end_time
+
+    end_header = ->(job, verbose:) do
+      job.end_time || verbose ? 'Ended at' : 'Estimated Finish'
```
benarmston, line 86: Let's have this as `Estimated end` to match up more clearly with `Ended at`.

benarmston (commented):
I'm not sure I understand how its collecting the estimated end time.  The monitor script seems to be missing a line.

Is it intentional that the estimated times remain set once the job has terminated?

```
[vagrant@chead1 (dev) flight-job]$ ruby ./bin/job info-job DKgN_Agm 
          ID: DKgN_Agm
   Script ID: WTYBE30m
Scheduler ID: 8
       State: COMPLETED
   Submitted: ✓
Submitted at: 02/06/21 14:58
  Started at: 02/06/21 15:13
    Ended at: 02/06/21 15:17
 Output Path: /home/vagrant/job-8.output
 Results Dir: /home/vagrant/simple.sh-outputs/8
[vagrant@chead1 (dev) flight-job]$ ruby ./bin/job info-job DKgN_Agm  --json
{
  "id": "DKgN_Agm",
  "created_at": "2021-06-02T14:58:28+00:00",
  "script_id": "WTYBE30m",
  "submit_status": 0,
  "submit_stdout": "{  \"id\": \"8\",  \"stdout\": \"/home/vagrant/job-8.output\",  \"stderr\": \"/home/vagrant/job-8.output\",  \"results_dir\": \"/home/vagrant/simple.sh-outputs/8\"}",
  "submit_stderr": "sbatch wrapper output:\nSubmitted batch job 8\n\nscontrol output:\nJobId=8 JobName=simple.sh\n   UserId=vagrant(1000) GroupId=vagrant(1000) MCS_label=N/A\n   Priority=4294901753 Nice=0 Account=(null) QOS=(null)\n   JobState=PENDING Reason=Priority Dependency=(null)\n   Requeue=1 Restarts=0 BatchFlag=1 Reboot=0 ExitCode=0:0\n   RunTime=00:00:00 TimeLimit=3-00:00:00 TimeMin=N/A\n   SubmitTime=2021-06-02T14:58:28 EligibleTime=2021-06-02T14:58:28\n   AccrueTime=Unknown\n   StartTime=Unknown EndTime=Unknown Deadline=N/A\n   SuspendTime=None SecsPreSuspend=0 LastSchedEval=2021-06-02T14:58:28\n   Partition=all AllocNode:Sid=chead1:10551\n   ReqNodeList=(null) ExcNodeList=(null)\n   NodeList=(null)\n   NumNodes=1 NumCPUs=1 NumTasks=1 CPUs/Task=1 ReqB:S:C:T=0:0:*:*\n   TRES=cpu=1,mem=1M,node=1,billing=1\n   Socks/Node=* NtasksPerN:B:S:C=0:0:*:* CoreSpec=*\n   MinCPUsNode=1 MinMemoryNode=1M MinTmpDiskNode=0\n   Features=(null) DelayBoot=00:00:00\n   OverSubscribe=OK Contiguous=0 Licenses=(null) Network=(null)\n   Command=/home/vagrant/.local/share/flight/job/scripts/WTYBE30m/simple.sh\n   WorkDir=/home/vagrant\n   StdErr=/home/vagrant/job-8.output\n   StdIn=/dev/null\n   StdOut=/home/vagrant/job-8.output\n   Power=\n   NtasksPerTRES:0\n",
  "state": "COMPLETED",
  "scheduler_id": "8",
  "stdout_path": "/home/vagrant/job-8.output",
  "stderr_path": "/home/vagrant/job-8.output",
  "results_dir": "/home/vagrant/simple.sh-outputs/8",
  "scheduler_state": "COMPLETED",
  "estimated_start_time": "2021-06-05T15:01:50+00:00",
  "estimated_end_time": "2021-06-08T15:01:50+00:00",
  "reason": null,
  "start_time": "2021-06-02T15:13:50+00:00",
  "end_time": "2021-06-02T15:17:51+00:00"
}
[vagrant@chead1 (dev) flight-job]$ 
```

Otherwise the functionality seems good.

libexec/slurm/monitor.sh
WilliamMcCumstie: Nope, that looks to be correct. See commit:

2403ddace18ef96a2b914a880355c18cad2d8ac8

WilliamMcCumstie: 5b98b81d85392e49a356e42a305e1e98ed90db59

lib/flight_job/outputs/info_job.rb
WilliamMcCumstie: 90209897cc99b8a7041c24755b5e82ad9ece3e4a

WilliamMcCumstie:
@benarmston In regards to how `estimated_{start,end}` are extracted from `slurm`, in short, they are guesses.

`slurm` does not distinguish between estimated and actual times. In either case, they come from the same `StartTime`/`EndTime` fields (when using `scontrol`).

The only way to distinguish the two, is two interrogate the current `state`. This is easier said than done due to the `state-map.yaml` file. For this reason, it is done within ruby to ensure the times are set in a consistent manner.

All the `monitor.sh` script needs to do is guess what the estimated/actual times are, and `ruby` will do the rest. This isn't ideal, but are options are limited when it comes to processing `scontrol` output.

---

In regards to "unsetting" the estimated times, I wouldn't know why we would want to? They are stored as separate fields from the actual times, so there is no need to unset it.

---

One way around the above two sections is to scrap the `estimated_start_time` and `estimated_end_time` fields entirely. They don't work particularly well for `slurm` anyway.

Instead the `monitor.sh` would report the `start_time`/ `end_time` regardless if the job is pending,running,terminal (when known). Then `ruby` can infer if it is the actual or estimated time from the `state`. Doing it this way has the following advantages:

* The `state` is already being used to infer estimated/actual times,
* There is no need to "unset" an estimated time on a `state` transition, and
* It is not forcing `slurm` to provide information it doesn't have.

benarmston:
What was confusing me was the combination of passing 1) setting `start_time` from `estimated_start_time` in the bash script itself; and 2) setting `estimated_end_time` from `end_time` in the JQ template.

I'd describe `slurm` as knowing full well the difference between a start time and an estimated start time, but not bothering to make that distinction clear to the user.  Instead the user needs to maintain their own mapping of the job states for which the start time is actually only an estimated start time.

I want `flight-job` / the web app to make the distinction clear to the user, without having to maintain a map of job states in multiple locations, e.g., `flight-job` and `flight-job-webapp`.

I don't want to scrap the `estimated_{start,end}_time` from `FlightJob::Job` or from the response returned by API.  However, if you are suggesting that we remove it from the communication between `FlightJob::Job` and `monitor.sh` that sounds like a good idea.

WilliamMcCumstie:
> What was confusing me was the combination of passing 1) setting `start_time` from `estimated_start_time` in the bash script itself; and 2) setting `estimated_end_time` from `end_time` in the JQ template.
> 
> I'd describe `slurm` as knowing full well the difference between a start time and an estimated start time, but not bothering to make that distinction clear to the user. Instead the user needs to maintain their own mapping of the job states for which the start time is actually only an estimated start time.
> 
> I want `flight-job` / the web app to make the distinction clear to the user, without having to maintain a map of job states in multiple locations, e.g., `flight-job` and `flight-job-webapp`.
> 
> I don't want to scrap the `estimated_{start,end}_time` from `FlightJob::Job` or from the response returned by API. However, if you are suggesting that we remove it from the communication between `FlightJob::Job` and `monitor.sh` that sounds like a good idea.

Sure thing, we can keep it in the presentation layer to the `user` and `flight-job-webapp`. That way they don't need knowledge of the mapping file.

Will then remove the estimated times from the `monitor.sh` script protocol and underlying metadata.

benarmston:
I'm surprised you're thinking of dropping it from the metadata.  I would have thought it would need to be there in order to avoid duplicating the mapping file in the API.  Either way, as long as it makes it easy for me to produce something like:

image

and

image

WilliamMcCumstie:
> I'm surprised you're thinking of dropping it from the metadata. I would have thought it would need to be there in order to avoid duplicating the mapping file in the API. Either way, as long as it makes it easy for me to produce something like:
> 
> image
> 
> and
> 
> image

There isn't perfect parity between the `--json` output and `metadata`, so dropping the key won't cause an issue.

More over, it side steps the question "Do we unset the estimated time once the actual is known?".

WilliamMcCumstie:
Rebased on master

WilliamMcCumstie:
On review we need to make a distinction within the `monitor.sh` script between `start_time` and `estimated_start_time`.

`slurm` will assign cancelled jobs a "start_time" even if they never actually started (for reasons). Unfortunately there are two categories of cancelled jobs:

* Ones which never started and should not have a `start_time`, and
* Ones which did start and do have a `start_time`.

The only way to distinguish between the two is to check if the job received an "allocation". This means the `monitor.sh` must be able to report `start_time` and `estimated_start_time` independently.

For consistency the same is done for `end_time`/`estimated_end_time`. This distinction is never used within the `slurm/monitor.sh` script, however the generation of the JQ template has been clarified.

---

From one of your original comments you mentioned the following:

> I'm not sure I understand how its collecting the estimated end time. The monitor script seems to be missing a line.
> 
> Is it intentional that the estimated times remain set once the job has terminated?
> 
> ```
> [vagrant@chead1 (dev) flight-job]$ ruby ./bin/job info-job DKgN_Agm 
>           ID: DKgN_Agm
>    Script ID: WTYBE30m
> Scheduler ID: 8
>        State: COMPLETED
>    Submitted: ✓
> Submitted at: 02/06/21 14:58
>   Started at: 02/06/21 15:13
>     Ended at: 02/06/21 15:17
>  Output Path: /home/vagrant/job-8.output
>  Results Dir: /home/vagrant/simple.sh-outputs/8

This implies you see the "start_time" and `estimated_start_time` as mutually exclusive. If this is the case, they should not be stored as separate fields in the `metadata`. Doing so allows the metadata to get into an inconsistent state. It would also violate the 2NF (if this was a db) by introducing a dependency between the `*-start-time` attributes and `state`.

It is far easier to store a single `start_time` field within the metadata and then infer the `actual_start_time` and `estimated_start_time` from the `state`. This way the two times are mutually exclusive by design.

NOTE: The above applies equally to `end_time`. 

This greatly simplifies the logic on how `start_time` and `end_time` is set when processing the `monitor.sh` output:

* `pending`: Both `start_time` and `end_time` are set from the estimated fields,
* `running`: The `start_time` is set from the actual field and `end_time` from estimated, and
* `terminal`: Both `start_time` and `end_time` are set from the actual fields.

---

As previously mentioned, the `--json` output is not a direct mapping of the `metadata` file. Four new keys have been added to it:

* `actual_start_time`
* `estimated_start_time`
* `actual_end_time`
* `estimated_end_time`

The corresponding API PR needs to be updated to use the `actual_*_time` attributes, but otherwise it is compatible with the change. It also prevents the API/user needing to know about the state mapping file.

NOTE: The `index-job` and `info-jobs` outputs have been updated to use the actual times where appropriate. So there specification is unchanged.

WilliamMcCumstie:
Note: I use the term "mutually exclusive" to refer to actual/estimated times. This is not *technically* true for the trivial case where they are both `nil`. This edge case has already been handled, making the times "pseudo-mutually exclusive".

PR #19, opened 2021-05-06
Add commands for displaying the standard output/error

WilliamMcCumstie:
~This PR is based on #17, which will need to be merged first.~

---

This has been implemented in two different ways depending on `combine_commands`. When `combine_commands` is `false`, they are separate commands:

```
flight job view-job-file JOB_ID FILENAME
flight job view-job-stdout JOB_ID
flight job view-job-stderr JOB_ID
```

If `combine_commands` is true, the are merged into a single command:

```
flight job view-job JOB_ID [FILENAME] [--stdout] [--stderr]
```

The underlying code base is the same for the two implementations, so there is a bit of redundancy. Once a CLI is selected, it can be simplified.

### DIFF
--- a/etc/state-maps/slurm.yaml
+++ b/etc/state-maps/slurm.yaml
@@ -70,7 +70,7 @@ RQ:             PENDING
 REQUEUED:       PENDING
 RS:             RUNNING
 RESIZING:       RUNNING
-RV:             FAILED # Currently jobs can not be tracked if they change cluster
+RV:             FAILED # Currently jobs cannot be tracked if they change cluster
 REVOKED:        FAILED # as above
 SI:             RUNNING
 SIGNALING:      RUNNING
--- a/lib/flight_job/cli.rb
+++ b/lib/flight_job/cli.rb
@@ -183,8 +183,35 @@ module FlightJob
       c.summary = 'Display details about a submitted job'
     end
 
+    # TODO: Remove all combine_command handling
+    signature = FlightJob.config.combine_commands ? 'JOB_ID [FILENAME]' : 'JOB_ID FILENAME'
     create_command 'view-job-output-file', 'JOB_ID FILENAME' do |c|
       c.summary = "View a file within the job's output directory"
+      if FlightJob.config.combine_commands
+        c.slop.bool '--stdout', 'Display the standard output'
+        c.slop.bool '--stderr', 'Display the standard error'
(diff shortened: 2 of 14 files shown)


benarmston:
@WilliamMcCumstie I like the use of `combine_commands` to switch between the competing CLIs.

I was expecting to be able to switch with the following:

```
flight_JOB_combine_commands=true flight job --help
```

However, that didn't work due to the following line https://github.com/openflighthpc/flight-job/blob/4275cb33b5a92938d302242397e209d8476c9d29/lib/flight_job/configuration.rb#L47.  I'm not sure why we want this and the comment above doesn't clarify.

benarmston:
I've rebased this on `master`.  The separate commands are definitely the better commands.  I'm going to see if I can think of a better way of handling `view-job-stderr` when stderr is merged with stdout.

Other than what's mentioned this is looking good.

benarmston:
I've added commits that implement my preferred spelling of the various commands.  We now have:

* `list-job-results JOB_ID`: list the contents of the job's results directory.
* `view-job-results JOB_ID FILE`: view the contents of a single file in the job's results directory.
* `view-job-stdout JOB_ID`:  view the job's standard output
* `view-job-stderr JOB_ID`: view the job's standard error (unless merged with standard output)

I've also renamed the bash script variable from `OUTPUT_DIR` to `RESULTS_DIR`.

This now looks good to me.  I'd still like your comments on the above issue I raised about environment variables, but that can take place separate to this PR.

WilliamMcCumstie:
> @WilliamMcCumstie I like the use of `combine_commands` to switch between the competing CLIs.
> 
> I was expecting to be able to switch with the following:
> 
> ```
> flight_JOB_combine_commands=true flight job --help
> ```
> 
> However, that didn't work due to the following line
> 
> https://github.com/openflighthpc/flight-job/blob/4275cb33b5a92938d302242397e209d8476c9d29/lib/flight_job/configuration.rb#L47
> 
> . I'm not sure why we want this and the comment above doesn't clarify.

The `env_var: false` has been around for awhile, so I forgot to mention it.

Environment variable overrides have been disabled by default for `flight-job` as it needs to be wrapped by the CLI. This is to prevent users from re-configuring the application within their `.bashrc`; which isn't sourced by the API.

This helps to ensure consistency between the two.

PR #21, opened 2021-05-21
Update flght-job to use the new configuration mechanism

WilliamMcCumstie:
This is broadly based on how `flight-cert` uses the new configuration mechanism but with a couple of key differences:

* It uses the bleeding edge of `FlightConfiguration` to provide validation support,
* The `Commander::ErrorHandler` is manually invoked to catch the previously mentioned validation  error,
* The dotenv files uses `.env{.local}` instead of `.env.development{.local}`, and
* The `root`/`logger`/`config` is defined on `Flight` directly.

---

## FlightConfiguration

This build is primarily used to test out the new validation features built into `FlightConfiguration`. However it means `FlightJob::Configuration.load` may raise an error.

This is to be expected and is not a major problem, however it means the config load needs to occur within the error handler provided by `Commander`.

The easiest way to achieve this is to manually invoke `Commander::ErrorHandler(name).start` instead of calling `CLI.run!`. The difference between `CLI.run!` and `CLI.run` is this error handler.

---

## Naming of dotenv files

The broad use of `dotenv` files is the same as `flight-cert`. On review however, I believe the `.development` extension is superfluous. This is mainly because:

* There is nothing stopping someone using `.env.development` in a production environment,
* In a standalone "alternative production" install it could be useful to force the environment, and
* In a rpm production install, it provides a mechanism to force a configuration which can not be modified.

Whilst I agree that the `.env` files generally shouldn't be used in `production`, I am hesitant to make that a hard condition. Instead the general usage of `,env` has been added to `bin/job`.

---

## Where `Flight` is defined

The `flight-cert` package defines `Flight` within the `lib/flight_cert.rb` file. Then `Flight` delegates to `FlightCert` to provide the interface.

This causes a bit of an issue within `FlightJob` as `lib/flight_job.rb` isn't required until a command is executed. This is done to prevent rogue `autoloads`/`requires` occurring before the help text is rendered.

Instead the `Flight` module is defined within `lib/flight_job/configuration`. Then `FlightJob` delegates to the `Flight` module.


benarmston:
Rebased on latest `master`.

benarmston:
I'll add all my review comments here as github is playing up at the moment.

* Defining `root`,`logger`,`config`,`env` directly on `Flight` is good.
* I've **added a commit** moving the `Flight` module to `lib/flight.rb`.  This allows access to `Flight.env` without loading the configuration.  Doing that allows `ENV['flight_ENVIRONMENT']` to be removed from `bin/job` in favour of `Flight.env.development?`.
* I've **added a commit** removing `Flight.config.development`.
* Regarding `run` vs `run!`.  It looks like we probably want a variant that takes a block.
    ```ruby
    def run_with_setup(*args, &block)
        Commander::ErrorHandler.new(program(:name)).start do |handler|
          block.call
          run(*handler.parse_trace(*args))
        end
    end
    ```

Regarding dotenv files:

* In general there is nothing wrong with using `.env` files in production.  We don't do that however.  For our services we have `/opt/flight/etc/service/env/` which serves the same purpose of a `.env` file.
* Our usage of dotenv files is purely a pragmatic method of enabling `/code/flight-job/bin/job` to run in development mode.  Whilst `flight job` runs in production mode.  It's simple and it works.
* I want an easy method to *communicate* that these dotenv files are only intended for development and not intended as part of production configuration.
* `.env.development` *communicates* that it is only for development.  If it ends up being included in an RPM it looks wrong and is likely to be picked up as being wrong much sooner than `.env` would be.
* If an admin is creating a `.env.development` or a `.env.development.local` file it is likely to *feel* wrong to them.  They are less likely to do so and more likely to look for another means of configuring the application.
* At some point we may remove the dotenv files if we come up with a better mechanism to have `/code/fligiht-job/bin/job` run in development mode.

The *technical* points you raise are all correct, however the use of `.env.development` is about *communicating intent* rather than any technical restriction.

WilliamMcCumstie:
In regards to the `run_with_setup` I'm not sure how that would work. Whilst I agree starting the `error_handler` manually is a bit of a pain, we don't have many choices.

The `bin` file setup is highly order dependent:

1. Firstly `Bundler` needs setting up and `commander` required (*),
2. `Commander` is used to start the `error_handler` (**),
3. The `config` is loaded within the error handler in case validation fails,
4. The `cli` is loaded which now has access to the `config`,
5. The command can be ran with `CLI#call`.

\* Dev dependencies are setup here as well, but this is somewhat unrelated
\** `--trace` handling should be started ASAP after this point

This means `run_with_setup` should not be defined on `FlightJob::CLI`. Doing so would prohibit the `config` being used within the CLI setup. Currently the `config` isn't used within `FlightJob::CLI` class setup, but has been in the past.

I am hesitant to introduce a hard requirement that `cli` load must occur before `config` load. This seems like an unnecessarily restriction for what is a minor code organisation change.

We _could_ define  `run_with_setup` method within the global namespace? However I don't know what this would achieve.

bin/job
```diff
@@ -35,10 +35,12 @@ begin
...
 
   # Limited use of dotenv to support setting flight_ENVIRONMENT=development.
-  # NOTE: The GitHub .env default is 'development', but the underlying default
+  # NOTE: The GitHub .env.development default is 'development', but the underlying default
   #       is 'production' if both env files are omitted.
   require 'dotenv'
-  dot_files = [ '../.env.local', '../.env' ].map { |f| File.expand_path(f, __dir__) }
+  dot_files = [ '../.env.development.local', '../.env.development' ].map do |file|
+    File.expand_path(file, __dir__)
+  end
```
benarmston, line 43: This is different than what I used in `flight-cert`; preventing `dotenv` from traversing up the filesystem to find `.env*` files.  I like it.

benarmston:
I've added some commits that 1) add some move validations; 2) remove the `flight-` prefix from the config files; and 3) update `flight_configuration` to `0.3.0`.

With those changes; this LGTM.  Thanks for your work on this @WilliamMcCumstie

bin/job
WilliamMcCumstie: :+1:

PR #22, opened 2021-05-21
Various changes to flight-job

WilliamMcCumstie:
* Exits 21 not 20 if `load_template` is used with a name,
* Always display the script id in the `list-jobs` output
* Don't display the submission output if it is successful, and
* Change `Created at` to `Submitted at` in the job outputs

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -48,7 +48,7 @@ GEM
       rexml
     method_source (1.0.0)
     minitest (5.14.4)
-    output_mode (1.5.0)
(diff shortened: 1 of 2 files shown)


benarmston:
I haven't got a clue what any of the output mode changes mean, but it does produce the results I was after.  The output for successful submissions is much nicer.

WilliamMcCumstie:
> I haven't got a clue what any of the output mode changes mean, but it does produce the results I was after. The output for successful submissions is much nicer.

Mostly define more templates

PR #24, opened 2021-05-24
Use the standard missing command handler

WilliamMcCumstie:
RONSEAL


benarmston:
That is much better. :-)

PR #27, opened 2021-07-16
Various changes to support multiple schedulers

WilliamMcCumstie:
This PR makes various changes to support multiple schedulers:

## Default Configs Dependency on Scheduler

Various configs depend on the scheduler to set the file paths (e.g. `submit_script_path`, `state_map_path`, etc..). To change the selected scheduler would require updating every config. This process is likely to be error prone, particularly on updates. Instead the "default function" can optionally be passed the `config` which is being loaded. This allows the paths access to the new `scheduler` setting.

The `ConfigurationPatches` module contains the code required to update `flight_configuration` to support this. It makes the following changes:

* Redefines the `load` method to a custom `load2`,
* Prevents the `defaults` hash being generated before all the other configs are loaded,
* Resolves the defaults within `*_before_type_cast`, and
* Preforms the transforms within the "attr_reader".

By resolving the defaults within `*_before_type_cast`, it allows the other configs to load first. A similar change was required for the transforms.

## Update various paths to match the builder

The above mechanism does not work well, if the builder redefines the paths. Instead the core paths have been updated to match the builder:

* `state_map_path` => "etc/job/state-maps/<config.scheduler>.yaml",
* `submit_script_path` => "libexec/job/<config.scheduler>/submit.sh"
* `templates_dir` => "usr/share/job"
* etc ...

---

## Split the templates into `workflow` and `directives`

Previously each template was monolithic and contained everything it needed to run. This makes it hard to write templates which work well with other schedulers.

As the `directives` are tightly coupled to the `scheduler`, they have been split into a different file: `directives.<scheduler>.erb`. This makes the new template directory structure:

```
[vagrant@chead1 (dev) job]$ pwd
/code/job/usr/share/job
[vagrant@chead1 (dev) job]$ tree simple
simple
├── directives.slurm.erb
├── metadata.yaml
└── workload.erb

0 directories, 3 files
```

The `directives.<scheduler>.erb` and `workload.erb` are currently rendered against the same question set.

\* NOTE: Originally `workload.erb` was named after the `script_name` within `metadata.yaml`. This has changed to complement  the new directives file. There is an automatic symlink which will update the legacy name to `workload.erb`.

---

## Core configuration values

The current `workload.erb` files reference various `SLURM_*` environment variables. Removing the environment variables would obfuscate the script, leaving them next to unusable. The `workload.erb` files need to support "scheduler environment variables" in some manner.

This has been achieved by adding a new `template_map_path` config. This is a set of key-value pairs which are provided to the renderer:

```
[vagrant@chead1 (dev) job]$ cat etc/job/template-maps/slurm.yaml 
...
results_dir:      "${WORKING_DIR}/${SLURM_JOB_NAME}-outputs/$SLURM_JOB_ID"
job_id:           SLURM_JOB_ID
job_name:         SLURM_JOB_NAME
...
```

This allows the core configs to reference the generic erb tag (`<%= core.job_id %>`) which is then rendered to the familiar variables: `SLURM_JOB_ID`.

This is only intended for soft-coupling on commonly used variables within the `workload.erb` files. Any highly scheduler specific code should be located in `directives.<scheduler>.erb`.

---

## Lazy rendering of scripts + caching

When the `script` is created, it renders the `workload.erb` to a plain `workload` file. This allows the user to edit the workload with `edit-script`. This prevents the user from editing the directives.

Currently, user editing of directives is not supported. This means `directives.<scheduler>.erb` does not get rendered independently against the `script`. Instead the directives will be re-rendered on the fly when required. This should make re-asking questions easier in the future.

There is a "cached file" which is stored against the `script`. The cache is updated each time the script is rendered. This is primarily so `flight-job-script-api` can quickly load the script without calling `flight job`.

As a cache file, it may go stale if the `workload` is manually edited. This is mitigated against in two ways:
* It is not used internally within `flight-job` in favour of re-rendering, and
* `create-script`, `info-script`, `view-script`, `edit-script`, and `submit-job` all trigger a re-render.

---

## Updated rendering + adapter script

When ever `flight-job` renders a script it concatenates three templates: `directives`, `adapter`, and `workload`. The new `adapter_script_path` contains `flight-job` related code that needs to be rendered into every script. Currently this is used to set the `RESULTS_DIR` according to the scheduler's convention.

`flight-job` is un-opinionated on how `results_dir` is set, instead it is entirely up to `submit.sh` to report the directory. Currently there is a convention between the `templates` and `submit.sh` that ensures consistency. Unfortunately, this convention is coupled to `slurm` and will need to be re-worked for another scheduler.

To make this easier, the `RESULTS_DIR` has been moved from `directives.<scheduler>.erb` into `adapter.<scheduler>.erb`. This un-couples the `templates` from `submit.sh`.

It also provides a location for future `flight-job` features to hook into the script, without updating the templates.

---

Other miscellaneous changes:

* Each `job` now stores a copy of their rendered "script file",
* `format_time` has been moved of `Job` and onto `Outputs`.


etc/job.yaml
```diff
@@ -48,45 +48,102 @@
...
+
+# ==============================================================================
+# Template Map Path
+# Specify the path the template map path. This should contain scheduler specific
+# environment variables that will be rendered into the templates.
+#
+# This allows the templates to be written in a platform agnostic manner, but
+# rendered with familiar variables.
+#
+# The default value will automatic use the 'scheduler' config. This feature
```
benarmston, line 122: Copy change: automatic -> automatically.  And similar for the other directives please.

benarmston (changes requested):
@WilliamMcCumstie the work done on separating the directives from the workload looks mostly good.  I'll take a further look at it once the changes noted below have been made.

However, an important part of this work was that there were to be no noticeable changes to `flight-job`.  From a user's perspective, the internal changes should not be observable.  This is important for an iterative improvement of the functionality without breaking any existing functionality.

Unfortunately, as you noted in the PR description, the way that the scripts are rendered this is not the case.  It is no longer possible to edit the scheduler directives once the script has been created.  It is important that we continue to support editing the directives through the `edit-script` command and the webapp.

Eventually, we will likely want to have the `edit-script` command only provide access to the workload.  However, that change needs to go hand-in-hand with an ability to re-ask the questions, and possibly an "advanced edit" mode in which the user gains access to the fully rendered script including the directives.

Could you split this PR into one which separates the directives without changing the behaviour of `edit-script` and a second one which adds the "delayed" rendering.  I think the delayed rendering is close enough to what we will eventually want that I don't want to lose it.

---

Also, this branch of flight job does not work with the webapp.  Both the latest `master` and the released `2.3.0` produce an unexpected error when trying to view a script, view its notes, or submit a script.

Please take a look at that too.

WilliamMcCumstie:
Closing in favour of a new set of PRs

PR #29, opened 2021-07-22
Standardise the config and update the README

WilliamMcCumstie:
This PR does two things:

* Standardise the `config` so it matches the `builder`, and
* Document the installation/configuration in the README.

---

The standardised config work is broadly the same as #27 is need to make supporting multiple schedulers easier. It also means the documentation applies to the `production` rpm build as well as a "standalone" usage.

It does mean a standalone install can not be seamlessly upgraded from and older version to the current one. This will *hopefully* be a one time annoyance. All future releases should have parity between the `builder` and standalone configs.

---

The documentation introduces the concept of a `standalone` environment. This has not been QA and is still at the "proof of concept" stage. However the existing configuration mechanism already implicitly supports environments other than `production`/`standalone`, this PR formalises the usage.

Broadly there are three environments:
* `production` - Respects `flight_ROOT` and uses the production gems,
* `standalone` - Ignores `flight_ROOT` but still uses the production gems, and
* `development` - Ignores `flight_ROOT` and uses the development gems.

By using a `standalone` environment, it makes the usage of `flight_ROOT` clearer.

Currently `production` will default to `standalone` behaviour if `flight_ROOT` is not set. Should this be changed to raising an error instead?

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,21 @@ To prevent this, the `run-monitor` command should be ran periodically for each u
 
 \*\* NOTE: Custom `scheduler` implementations should also transition missing jobs to an `UNKNOWN` state. Failure to do so will cause them to get stuck in the "last known state".
 
+# Versioning
+
+This application uses [semver](https://semver.org/) to version the command line interface. The current version is given within [version.rb](lib/flight_job/version.rb). This version is limited to the applications _public command line interface _(exhaustive list):
+
+* The command names and associated aliases,
+* The commands' type signatures,
+* The commands' option flags (unless otherwise stated),
+* The column order of the machine readable index outputs,
+* The key names of the machine readable show outputs,
+* The required keys within the JSON outputs (\*\* pending specification)
+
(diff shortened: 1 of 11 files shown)


WilliamMcCumstie:
A couple more commits have been added which extract the `monitor.sh` and `submit.sh` response specifications into dedicated files.

These specifications are necessary to write custom scheduler scripts. Therefore it is handy to be able to link against them in the `README.md`

WilliamMcCumstie:
Also flag that these are the `v0` versions of the specifications.

WilliamMcCumstie:
Rebased on master

WilliamMcCumstie:
Rebased on master

benarmston:
I've made changes to the README.  Mostly reducing its length by extracting sections to a new `docs/` directory.  I've also expanded on custom template creation, scheduler integrations and updated to the operations section.

I'm sure that this will quickly become outdated, but LGTM right now.

PR #30, opened 2021-07-22
Initial work to support interactive job submissions

WilliamMcCumstie:
This PR contains the initial work to allow jobs to be submitted interactively and connected to with `flight desktop`. It does the following:

* Introduces a `controls_dir` to contain the output of `flight-desktop`,
* Adds `simple-interactive` template to demo an interactive job,
* Adds the `time` question type, and
* Initial support for "start deadlines" with the `--deadline` flag.

---

## Controls Dir

The `controls_dir` behaviour has been broadly based on how `results_dir` works. The difference between the two is one of purpose:

1. `results_dir` is user facing and contains the various output files their job may generate, where
2. `controls_dir` is an internal directory to be used by flight-job.

When an interactive job runs, it needs to run a `flight-desktop` command and redirect its output to:
* `stdout` => `$CONTROLS_DIR/flight-desktop-stdout`
* `stderr` => `$CONTROLS_DIR/flight-desktop-stderr`
* `status` => `$CONTROLS_DIR"/flight-desktop-status`

By writing these special files to a `$CONTROLS_DIR`, it keeps them hidden from the end user. However it also allows `flight-job` indirect access to `flight-desktop`.

Note: Currently the internal implementation is rather rudimentary and only uses `flight-desktop-stdout`. However this will likely need to be improved before final release.

---

## Simple Interactive Template

There is a new `simple-interactive` template which is broadly based on the `simple` template. It uses `gridware` to launch an `R` session. The workload section is fairly straight forward as it is based on other work.

Unfortunately it doesn't currently work. It relies on the new `flight desktop start --app` flag which has not been released. I am not keen on trying to switch out `flight desktop` to a dev version. Instead the template will remain broken until `flight-desktop` is released.

---

## Time question type

A new `time` question type has been added. This is designed to accept a time of day in the `HH:MM` format. This is design to correspond with a `<input type="time">` HTML tag.

This does create a conflict with the existing SLURM time formats. The `max_runtime` input accepts multiple SLURM time formats. This creates a few issues:

* It does not port well to other schedulers,
* The user needs to be familiar with slurm's esoteric time formats, and
* It uses the MM:SS and HH:MM:SS formats.

This results in the questions using both the `HH:MM` and `MM:SS` time formats. Even with documentation, this is highly confusing and will need to be addressed.

---

## Start Deadlines

As these are interactive jobs, they tend to have priority on the scheduler's queue with short maximum runtimes. This is to allow the end user to get on/off the nodes quickly.

This creates the need to set a deadline in which the job needs to _start_. Three deadline types have been implemented:

* `immediately`: A slight misnomer, the job needs to start within 5 seconds of submission,
* `today`: The user providers a time in which the job must start on the same day, and
* `offset`: The user specifies the maximum time they are willing to wait for the job to start.

The three types have all been implemented with slurm's `--deadline=now+<seconds>` directive. Unfortunately `--deadline` uses the "predicted end time" to enforce the deadline. This means the `--deadline` needs to be offset by the `max_runtime` of the script, regardless of which option above is selected.

NOTES:

The `--begin` is not used as this delays the start script (AKA do not start before this time),

The `--deadline=HH:MM` is not use as it has different behaviour depending if:
* If `start_time`/`deadline` is the future, schedule the job with those times,
* If `start_time`/`deadline` is in the past, add 24 hours and schedule the job,
* If `start_time` is in the past, but `deadline` in the future, then do not add 24 hours and fail to schedule the job as `max_runtime` exceeds the `deadline`.

Where `start_time` is the provide time of day from the user, and `deadline` is `start_time` plus the `max_runtime`.

As a side effect of the above constraints, the `--deadline` mechanism assumes the `job` is submitted immediately after rendering the `script`. This is currently not the case, so the mechanism will not function correctly. However rendering the `directives` dynamically at submission time is already in the pipeline and will fix this issue.


WilliamMcCumstie:
Also adds a `--wait-flag` to `info-job`. This causes the command to hang until either the job terminates or reports it's `flight-desktop` session.

benarmston (approved):
As discussed IRL, I've made quite a few changes here.

### Issues with the initial implementation:

The interactive job script was setting the environment variables in a way that could not necessarily be relied upon to be available to the desktop session.  To make sure that they are available, a session script is created and passed to flight desktop via the `--script` option.

Also, the job script was not terminating when the session was ended.  The initial implementation used `sleep infinity` to sleep until the scheduler killed the job.  Now the job script periodically checks if the session is still running and if it isn't it quits.

### Controls directory and files

I've expanded on the concept of a controls directory and controls files.  I've added `ControlsDir` and `ControlsFile` classes to encapsulate common access patterns to them.

The controls directory is now located in the job's directory.  E.g., `~/.local/share/flight/job/jobs/<JOB ID>/controls`.  This path is hardcoded as I can't see any reason that we'd want to configure it.

The entire contents of the controls directory is serialized when the `--json` flag is given.

The job script now writes the `flight-desktop-id` and `flight-desktop-status` to the controls directory.  The `flight desktop start` stdout and stderr is sent to the job script's stdout and stderr.  This has proved very useful in debugging issues I've had.

Parts of the session script output are also redirected to `session.output` in the results directory.  This is expected to be useful in debugging issues with session scripts.

### Other changes

I've removed the question about the deadline (aka the latest time at which the job can start).  In hindsight, until we have submission time questions, asking this question doesn't really make sense.  I suspect that much of the work you've done will be resurrected at that point.

I've added a new `tags` attribute to templates and scripts.  Each tag is expected to be in the tripple tag format.  Each template has a tag of `script:type=batch` or `script:type=interactive`.  Interactive templates, also have a `session:type=desktop` tag.  The `script:type` tag replaces the previous `interactive` metadata.

When a script is created, the tags are copied from the template to the script.  This is not done when a job is created, but we may find a reason to want that in the future.

These tags are currently used to control how the client interprets the templates/script/jobs.  Later, the adapter mechanism may use these tags if different adapters are required for interactive vs batch scripts.

I've also removed the `--wait-desktop` flag.  I like the idea, but in practice it wasn't a useful way for the client to be informed of when the desktop session became available.  We now rely on the job page being "refreshed" every minute.  This will be sufficient for now.

PR #31, opened 2021-08-03
Warn if re-ask selected questions is empty

WilliamMcCumstie:
It is fairly easy for a user to exit out of the "re-ask selected questions" prompt without selecting any questions. This needs to be _allowed_ otherwise there wouldn't be an easy way to exit the prompt. 

Instead a warning prompt is issued before the user is redirected to the main REPL. This warning tells the user they need to select questions with `Space`. The main REPL defaults to the "re-ask selected questions" prompt.

### DIFF
--- a/lib/flight_job/commands/create_script.rb
+++ b/lib/flight_job/commands/create_script.rb
@@ -199,6 +199,7 @@ module FlightJob
         def prompt_again
           opts = {
             default: if @prompt_for_selected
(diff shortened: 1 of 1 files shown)


benarmston:
Rebased on `master`.  Prompt formatting adjusted slightly.

benarmston:
LGTM

PR #32, opened 2021-08-05
Replace the --include flag with flight_JOB_includes env var

WilliamMcCumstie:
The `--include` feature is only required by the API and shouldn't be part of the CLI's public interface.

Instead, it has been replaced with a `flight_JOB_includes` config/env var. This config has been documented as volatile and subject to change. This is to allow time for its future to become clear.

### DIFF
--- a/etc/job.yaml
+++ b/etc/job.yaml
@@ -139,6 +139,16 @@
 # submission_period: 3600
 
 # ==============================================================================
+# Include Related Resources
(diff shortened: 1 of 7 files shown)


benarmston:
I've removed mention of the `includes` mechanism from the config file as I don't think that there is a valid global value.  Instead it is now an "undocumented feature".

I've also rebased on latest `master`.  This LGTM now.

PR #33, opened 2021-08-11
Implement transient config dependencies on the selected scheduler

WilliamMcCumstie:
This PR is the first in a series to make supporting multiple schedulers easier in `flight-job`.

Key feature it implements is "transient" config dependencies. Previously this wasn't possible as the `configs` defaults where applies in the order the `__sources__` struct is loaded. Instead the following changes have been made:

* It uses the bleeding edge version  of `flight_configuration` which fixes  logging/validation,
* It delays resolving the defaults until the `source.value` is called by (`_before_type_cat`), and
* The top level `defaults` hash has been temporarily disabled.

* ~The `defaults` are intentionally omitted when `__sources__` is loaded, instead the proc is tracked,~
* ~`*_before_type_cast` has been redefined to either fetch the value from `__source__` or call the default proc,~
* ~The default proc will be passed the `config` object (respecting arity),~
* ~The newly transformed value are stored within a `__transformed__` struct, and~
* ~The "attr_reader" methods read from the `__transformed__` struct.~

Because the `defaults` are resolved on the fly when `*_before_type_cast` is called, the is no order dependency.

Whilst it is _technically_ possible to create a loop between the defaults, this is easily avoided by conservative use of transient defaults. Any situation which could create a loop would be an inappropriate use of the mechanism.

~*NOTE:* This has been implemented via a patch which redefines the various methods. This would need to be more formally integrated into `flight_configuration`.~

~*NOTE:* The use of a `__transformed__` struct allows for `nil`/`false` values to be cached~

---

~Currently the "attr_writer" methods have been dropped as `flight-job` does not use them. The use of "attr-writer" methods was previously a bit janky as there was no easy way to validate the new configs are correct.~

~Possible a better way would be to store them as a new "updated top level config". This would then be super-imposed over the existing `__sources__`/`__transformed__` structs. This should hopefully be a relatively minor change to implemented.~

---

Miscellaneous:

* The `format_time` method has been extracted onto `Outputs` to reduce duplication, and
* The various config paths have been updated to match the builder


benarmston:
I've 

1. rebased on latest `master`.
2. Merged the patches to flight configuration to flight configuration
3. Added explicit `presence` validations in addition to the one added by flight configuration.
4. Fixed the documented submit.sh script path.

This now looks good to me.

PR #34, opened 2021-08-11
Split the workload and directives sections in the templates

WilliamMcCumstie:
This PR is based on #33, please merge it first and rebase

---

This PR focuses on making the `templates` scheduler agnostic. This is achieved with a few different mechanisms:

* The directives have been split into their own scheduler specific `erb` file,
* The workloads have been generalised with a `template-map`, and
* `flight-job` specific code has been split into a scheduler specific `adapter.erb` file.

---

## Splitting the Directives

The directives are now stored as a `<template>/directives.<scheduler>.erb` file. Because the directives are coupled to the scheduler, it is easier to switch out that fragment of the template entirely, then try to implement the logic within `erb`.

This means the directives will need to be re-implemented for each scheduler that is intended to be supported.

The directive files are optional and maybe omitted. This is to allow for legacy monolithic templates. These templates are not rendered with the adapter as doing so would break the directives section.

---

## Workload Section - Renaming

Original the `erb` template was named after the `script_name` within the `template`'s metadata. This design no longer makes sense as the directives are being stored independently. It makes renaming templates/scripts tricky due to the dependency.

Instead the following has been implemented:

* `templates` now have a `directives.slurm.erb` and `worload.erb`,
* `script` now have a rendered `script.sh`,
* `jobs` copy the rendered script into their directory as `<script-name>.sh`

There is a migration that runs when a `template` or `script` is loaded, which will symlink the legacy paths into the new locations.

---

## Workload Section - Template Map

~The workload sections tend to need access to scheduler specific variables, e.g.:~
* `SLURM_JOB_NAME`
* `SLURM_ARRAY_JOB_ID`
* `SLURM_ARRAY_TASK_MAX`
* etc ...

~It is not practical to completely abstract these variables from the `workload.erb` for two reasons:~

* ~There are a lot of variables that _could_ be used, and~
* ~The end user will be expecting to see these when they edit their scripts.~

~To get around this, a new `template-maps` config has been added. This mapping file is mounted into the renderer as `core.<key>` and is designed to be used in the workload templates.~

~This allows the `workload.erb` file to use generic tags (e.g. `<%= core.job_id %>`), which are rendered to scheduler variables within the scripts (`SLURM_JOB_ID`).~

~This mapping file is intention non-exhaustive as the required key-maps are known in advance. Trying to map every `SLURM` variable will:~

* ~Loosely couple the `template-map` to slurm and reducing its utility, and~
* ~Likely be error prone and unnecessarily verbose.~

~*NOTE:* Custom templates are allowed (expected?) to hard code scheduler specific concepts into their workloads. It is probably easier to hard couple custom templates to the scheduler they where designed for.~

EDIT: This feature has been removed. The templates' `workload.erb` files still contain slurm specific environment variables. This will need to be addressed in a future PR.

---

## Adapter Template

The last major component is an `adapter` erb template which contains how `RESULTS_DIR` is set. This will be rendered in between the `directives` and `workload` sections of the script.

This provides a single location that `flight-job` specific code can be injected into the script. This will enable future features to be rolled out to all existing `templates` without having to update the `directives`. It will also help keep the section separate from the `workload` once delayed rendering is implemented.


WilliamMcCumstie:
Rebased on the new version of transient configs

WilliamMcCumstie:
Rebased on master

WilliamMcCumstie:
Rebased on master - again

lib/flight_job/models/job.rb
```diff
@@ -346,17 +358,18 @@ def submit
...
-        script.script_path
-      ]
-
       # Generate the initial metadata path file
       FileUtils.mkdir_p File.dirname(initial_metadata_path)
       File.write initial_metadata_path, YAML.dump(active_metadata)
 
+      # Duplicate the script into the job's directory
+      # NOTE: Eventually this should probably be named after the job_name question
+      metadata["rendered_path"] = File.join(FlightJob.config.jobs_dir, id, script.script_name)
```
benarmston, line 367: We now have a `job_dir` method so this can be `File.join(job_dir, script.script_name)`.

usr/share/job/adapter.slurm.erb
```diff
@@ -0,0 +1,11 @@
...
+#  >>>> FLIGHT JOB ADAPTER
+# ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░
+
+#=====================
+#  Results directory
+#---------------------
+# By convention, job's submitted with `flight-job` are expected to save their
+# results to RESULTS_DIR.  `flight-job` provides easy access to any files
+# saved there.
+RESULTS_DIR="$(pwd)/${SLURM_JOB_NAME}-outputs/$SLURM_JOB_ID"
```
benarmston, line 11: This isn't the correct path for array jobs.  It was previously using `RESULTS_DIR="$(pwd)/${SLURM_JOB_NAME}-outputs/$SLURM_ARRAY_JOB_ID"`

Perhaps adding a tag to the array template and branching on that here would work?

benarmston (changes requested):
@WilliamMcCumstie I've ran into a few issues with this. Could you take a look at them please.


There are some errors which may have been introduced in this PR or a previous one.

```
[vagrant@chead1 flight-job]$ /code/flight-job-script-service/api/libexec/job list-templates  | cat
bin/job: uninitialized constant OutputMode::Outputs::Delimited
```

Similarly for `list-scripts` and `list-jobs`.  Can you take a look at those please.

---

The initial metadata path doesn't look correct.

```
[vagrant@chead1 (dev.cluster) flight-job]$ find var/jobs/jOBf6SB7/
var/jobs/jOBf6SB7/
var/jobs/jOBf6SB7/simple.sh
var/jobs/jOBf6SB7/metadata.yaml
var/jobs/jOBf6SB7/jOBf6SB7
var/jobs/jOBf6SB7/jOBf6SB7/metadata.initial.yaml
```

The duplicated job id looks very odd.


---

I've added a commit to fix an issue with determining the results directory for array jobs.  This was an existing issue.  Unfortunately, the adapter script has introduced a bug with array jobs.

lib/flight_job/renderer.rb
```diff
@@ -44,59 +71,61 @@ def default
...
-        end
-        DefaultsOpenStruct.new(questions) do |h, k|
-          question = Question.new(id: k)
-          h[k] = AnswerDecorator.new(question: question, answer: nil)
-        end
-      end
+    private
+
+    def render_adapter
+      # NOTE: The adapter is designed augment the directives,
```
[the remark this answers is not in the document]
WilliamMcCumstie: 19f69f727f70bc859204ace40138bbff96d4fdfd

lib/flight_job/models/job.rb
```diff
@@ -110,8 +121,8 @@ def self.load_all
         if job.valid?(:load)
           job.tap(&:monitor)
         else
-          FlightJob.logger.error("Failed to load missing/invalid job: #{id}")
-          FlightJob.logger.debug(job.errors)
+          FlightJob.logger.error("Failed to load missing/invalid script: #{id}")
```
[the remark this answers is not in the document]
WilliamMcCumstie: 19f69f727f70bc859204ace40138bbff96d4fdfd

lib/flight_job/models/job.rb
WilliamMcCumstie: 119deea06f38bd9857c38de081e30bf0a5f6ba46

WilliamMcCumstie:
The `initial_metadata_path` issues has been fixed in: 119deea06f38bd9857c38de081e30bf0a5f6ba46

WilliamMcCumstie:
I can not replicate the `OutputMode::Outputs::Delimited` issue:

```
[vagrant@chead1 job]$ /opt/flight/bin/ruby bin/job list-templates  | cat
1       interactive-desktop     /code/job/usr/share/job/templates/interactive-desktop/workload.erb
2       simple  /code/job/usr/share/job/templates/simple/workload.erb
3       simple-array    /code/job/usr/share/job/templates/simple-array/workload.erb
4       mpi-slots       /code/job/usr/share/job/templates/mpi-slots/workload.erb
5       mpi-nodes       /code/job/usr/share/job/templates/mpi-nodes/workload.erb
6       smp     /code/job/usr/share/job/templates/smp/workload.erb
```

---

I notice you are using a `libexec` wrapper script:
`/code/flight-job-script-service/api/libexec/job`

However `flight-job-script-service` does not have a `libexec` directory. Instead it is configured within `flight_JOB_SCRIPT_SERVICE_flight_job`: 

```
      {
        name: 'flight_job',
        env_var: true,
        default: File.join(ENV.fetch('flight_ROOT', '/opt/flight'), 'bin/flight job'),
        transform: ->(value) { value.split(' ') }
      }
```

I suspect this is an issue with your development environment.

usr/share/job/adapter.slurm.erb
WilliamMcCumstie: Using `tags` might be good long term solution, but the quick fix is to toggle on the env var:

```
if [ -n "$SLURM_ARRAY_JOB_ID" ]; then
  RESULTS_DIR="$(pwd)/${SLURM_JOB_NAME}-outputs/$SLURM_ARRAY_JOB_ID"
else
  RESULTS_DIR="$(pwd)/${SLURM_JOB_NAME}-outputs/$SLURM_JOB_ID"
fi
```

4fbe7c15f8a32ab81a5336d25c9ebdba5780d152

benarmston:
I don't think its anything to do with the libexec wrapper.  It's just a convenient place where I can toggle the API between the dev and RPM versions of flight-job.

```
[vagrant@chead1 flight-job]$ cat /code/flight-job-script-service/api/libexec/job 
#!/bin/bash

#export flight_JOB_log_path=/dev/stderr
#export flight_JOB_log_level=debug
/opt/flight/bin/ruby /code/flight-job/bin/job "$@"
[vagrant@chead1 flight-job]$ 
[vagrant@chead1 flight-job]$ 
[vagrant@chead1 flight-job]$ 
[vagrant@chead1 flight-job]$ /opt/flight/bin/ruby  /code/flight-job/bin/job list-templates
┌───────┬─────────────────────┐
│ Index │ Name                │
├───────┼─────────────────────┤
│ 1     │ interactive-desktop │
│ 2     │ simple              │
│ 3     │ simple-array        │
│ 4     │ mpi-slots           │
│ 5     │ mpi-nodes           │
│ 6     │ smp                 │
└───────┴─────────────────────┘
[vagrant@chead1 flight-job]$ /opt/flight/bin/ruby  /code/flight-job/bin/job list-templates | cat
bin/job: uninitialized constant OutputMode::Outputs::Delimited
[vagrant@chead1 flight-job]$ 
```

You'll notice above that I get the exact same output when running the binary more directly.

benarmston:
I've updated the interactive desktop template to have a separate directives section too.  This required the adapter to be able to distinguish between interactive and non-interactive templates, so the template has been exposed to the render environment.  With that done, setting `RESULTS_DIR` has been updated to branch on a template tag.

PR #36, opened 2021-08-18
Do not force array jobs to report an incorrect stdout/stderr

WilliamMcCumstie:
Previously all `job` submissions had to report back a `stdout`/`stderr`. Technically it could have been empty string, but this is more an oversight then part of the specification.

This causes issues with array jobs for a few reasons:
* `scontrol` reports the wrong paths for the main "job",
* The "main" job does not really have a `stdout`/`stderr` path,
* The paths for each "array task" does not become available until the individual job starts, and
* The current specification only supports a single set of paths.

---

As a stopgap measure, the `stdout`/`stderr` inputs are now optional on `submit.sh`. They will default to `/dev/null` if they are omitted. Ideally they would allow to be `nil`/`null`, however this is not currently possible. Both fields are part of the CLI public interface as a required key via the `--json` flag.

This means `array` jobs aren't flat out wrong when reporting back the paths. The next PR will allow them to update the `stdout`/`stderr` files on `monitor.sh`

### DIFF
--- a/lib/flight_job/cli.rb
+++ b/lib/flight_job/cli.rb
@@ -75,6 +75,10 @@ module FlightJob
     global_slop.bool '--ascii', 'Display a simplified version of the output, when supported'
     global_slop.bool '--json', 'Display a JSON version of the output, when supported'
 
+    # NOTE: NEXT MAJOR CLI RELEASE
+    # Please review the Outputs and <Model>#serializable_hash methods on the next major release
+    # These will need simplifying
+
     create_command 'list-templates' do |c|
       c.summary = 'List available templates'
     end
--- a/lib/flight_job/models/job.rb
+++ b/lib/flight_job/models/job.rb
@@ -301,10 +301,6 @@ module FlightJob
       stdout_path == stderr_path
     end
 
-    # NOTE: NEXT MAJOR CLI RELEASE
-    # * stdout/stderr - Should become optional as array jobs don't really have a "single"
-    #                   out/err file. Currently they get set to /dev/null as a stop gap.
-    #                   More accurately it should be null.
     def serializable_hash(opts = nil)
       opts ||= {}
       {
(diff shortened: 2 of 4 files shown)


lib/flight_job/cli.rb
```diff
@@ -75,6 +75,10 @@ def self.create_command(name, args_str = '')
     global_slop.bool '--ascii', 'Display a simplified version of the output, when supported'
     global_slop.bool '--json', 'Display a JSON version of the output, when supported'
 
+    # NOTE: NEXT MAJOR CLI RELEASE
+    # Please review the Outputs and <Model>#serializable_hash methods on the next major release
+    # These will need simplifying
+
```
benarmston, line 81: Can you explain this please?

WilliamMcCumstie:
Rebased on master

lib/flight_job/models/job.rb
```diff
@@ -390,8 +394,8 @@ def submit
         # Parse stdout on successful commands
         process_output('submit', status, out) do |data|
           self.scheduler_id = data['id']
-          self.stdout_path = data['stdout']
-          self.stderr_path = data['stderr']
+          self.stdout_path = data['stdout'].blank? ? "/dev/null" : data['stdout']
+          self.stderr_path = data['stderr'].blank? ? "/dev/null" : data['stderr']
```
benarmston, line 398: I presume something breaks if we use `nil` instead of `/dev/nuill`.  I think I'd prefer the UI to state that the stdout wasn't available rather than this

image

I can handle that client-side if its too awkward to do here.

benarmston (commented):
A couple of questions for you to look at please @WilliamMcCumstie

lib/flight_job/cli.rb
WilliamMcCumstie: Just a reminder that the `cli.rb` isn't the only file that makes up the public interface for `semvar` purposes.

If and when we look at making a major version bump, we have the opportunity to:
* Reorder the "human" and "machine" outputs to be consistent, and
* Modify the `keys` which appear in the `--json` output.

---

This is not to say we _should_ make those changes on a major version bump, but we _could_ do so.

benarmston: Instead of comments like this littered about the source code, I'd suggest a document listing what changes we consider acceptable (or unacceptable) for each of the semvar major, minor, patch releases.  I'll add such a document as part of the README consistency work.

lib/flight_job/models/job.rb
WilliamMcCumstie: My mistake, I am starting to loose track of `required` keys within the `--json` output. On review, the `stdout_path`/`stderr_path` are actually "recommended":

30a64c8f807c41840a0da13245c9387029e7a1a8

benarmston: Thanks for that @WilliamMcCumstie .  We now have the following which is closer to what has happened.

image

benarmston:
LGTM.

PR #37, opened 2021-08-20
Switch to using user_configs to set log_level and path

WilliamMcCumstie:
The user can now set the `log_level` in their user config


lib/flight_job/configuration.rb
```diff
@@ -36,12 +36,14 @@ module FlightJob
...
 
   class Configuration
-    extend FlightConfiguration::DSL
+    include FlightConfiguration::DSL
     include FlightConfiguration::RichActiveValidationErrorMessage
     include ActiveModel::Validations
 
     application_name 'job'
 
+    user_configs :log_level, :log_path
```
benarmston, line 45: `jobs_dir` and `scripts_dir` should be added here too.

benarmston:
I've decided that `temlates_dir` and `min_terminal_width` are also to be user-configurable.  This LGTM.

PR #39, opened 2021-08-26
Decouple the job's metadata from its outputs

WilliamMcCumstie:
Currently a `job's` output (in particular `--json`) is couple to the `metadata` syntax. This prevents any major refactoring of the `metadata` as it would introduce a breaking change in the CLI.

A new `JobDecorator` has been implemented with ~`Draper`~ custom class. It does the following:

* Contains the methods primarily responsible for _representing_ the `job`,
* Contains the JSON serializer methods for the `--json` flag, and
* Decouples the `metadata` from the `json` representation.

A `delegate_metadata` helper method is used to pull the `metadata`. This is a temporary stop gap measure whilst there is a single `metadata` specification. This will need refactoring as specifications are added in the future.


benarmston:
LGTM

PR #40, opened 2021-08-27
Preform the state mapping within monitor.sh

WilliamMcCumstie:
Based on #39 

The `state-map` file is no longer a core feature. This is to decouple `flight-job` from the scheduler implementation. Instead the `monitor,sh` needs to return two keys:

* `state` - The mapped `flight-job` state, and
* `scheduler_state` - The state used by the scheduler.

It also allows the `monitor.sh` script to set the `start_time`, `end_time`, `estimated_start_time`, and `estimated_end_time`. `flight-job` will no longer infer whether a time has been set correctly. Instead it will validate the times based on the `state`.

---

A new `libexec/slurm/parser.sh` has been added to contain the `scontrol`/`acct` parsera. This will be used in a future PR.


benarmston:
I've rebased on latest `master` and fixed the `ls-job-results` command.  It now LGTM.

PR #41, opened 2021-08-31
Split the job's metadata formats

WilliamMcCumstie:
Based on #40

---

The `job` metadata has been split into three separate formats:

* `INITIALIZING`.
* `SINGELTON`, and
* `FAILED_SUBMISSION`

The format is saved as the `job_type` key within the `metadata`. A two step validation occurs on `valid? :load`:

* The shared "required" keys are validated with `additionalProperties: true` set, and
* The sub-schema for the specific `job_type` is validated with `additionalProperties: false`.

_NOTE:_ Technically this could have been done as a single schema with the `oneOf` validatior. However this made the error messages highly cryptic.

---

## INITIALIZING

All jobs are created within the `INITIALIZING` state before they are submitted to the scheduler. This replaces the `initial_metadata_path` file. These `INITIALIZING` jobs are filtered out from `list-jobs` and error on `info-job`.

---

## SINGLETON

The `SINGLETON` type represent a non-array job which was correctly submitted to the scheduler. They could be a `batch` or `interactive` job.

Jobs are transitioned to `SINGLETON`/`pending` after `submit.sh` runs. Then they are managed by `monitor.sh` as before.

---

## FAILED_SUBMISSION

These jobs flag something went wrong during the submission process. This indicates one of the following occurred:

1. The `submit.sh` script failed,
2. `flight-job` exited before writing the metadata (e.g. interrupt/internal error), or
3. The `submit.sh` exited 0 but failed the JSON response validation

The `run-monitor`/`list-jobs`/`info-job` commands will transition `jobs` which fall into case 2 above. This occurs via a modified version of the old `transition_inactive` mechanism. These jobs will receive an exit code of 126.

Non-zero exit codes  (case 1) are saved directly to the metadata after the `submit.sh` script runs.

There is an edge condition (case 3) where `submit.sh` exits 0 but the JSON response otherwise fails validation. The exit status is modified to be `126` to reflect an error has occurred with an unknown status. The log files still record the original exit 0.

---

# JobTransitions

To reduce the size of `job.rb`, the majority of the transition related code has been extracted into `job_transitions/*_transitions.rb`. This improves the readability of the `job.rb` file.

---

# Active File

A new `active_index_file` is being used to locate the "active jobs". Active jobs are defined as:

* All `INITIALIZING` jobs,
* `pending/`running` `SINGLETON` jobs

These `active_findex_file` are used by the `run-monitor` command to locate the jobs. The index files are created along side the `metadata` and managed when `job` transition states.

The index files are also validated on `job.valid? :load`.


benarmston:
This PR breaks submission of array jobs.  I suspect that this will be fixed in one of the next PRs so I'll merge this as it is.

PR #42, opened 2021-09-02
Update submit.sh and add monitor-array.sh

WilliamMcCumstie:
This PR is based on #41 

---

This PR does the following:

* Add's the `ARRAY` job metadata syntax,
* Creates the array `Task` object,
* Updates the response syntax for `submit.sh` and
* Adds the `monitor-array.sh`

---

## ARRAY job metadata syntax

The new `ARRAY` job metadata syntax is slightly different to that of a `SINGLETON` job. This is because an array "job" does not _technically_ get ran by the scheduler. Instead, the scheduler will spawn array `tasks`.

This has the following implications for the array `job`:
* It does not have a _direct_ `start_time`/`end_time`,
* The `estimated_start_time`/`estimated_end_time` are slight misnomers,
* It does not have a `scheduler_state`, and
* It does not have a _direct_ `state`.

Most of the above attributes can be inferred from the `tasks`. Doing this will be implemented in a future PR.

---

## Array Tasks

An array `job` should start many `tasks`, which are saved as separate metadata files:

```
<lib-dir>/jobs/<job-id>/tasks/metadata.<index>.yaml
```

The metadata syntax is broadly based of a `SINGLETON` job type. The major differences being `tasks`:

* Do not have the `submit_*` attributes, and
* They must be created with a `scheduler_state`.

Displaying the tasks through the CLI will be implemented in a future PR.

---

## Updating Submit.sh

The `submit.sh` script must now report back if the job is a `SINGLETON` or an `ARRAY` job. This will trigger the job type being updated accordingly.

The `submit.sh` no longer reports back the `stdout`/`stderr` paths. This done for two reasons:

* Array `TASKS` must be created via the `monitor-array.sh` script, and
* Ideally there should be parity between `monitor.sh`/`monitor-array.sh` where possible.

The appropriate monitor script is now ran immediately after `submit.sh`. This should ensure the paths are set.

---

## Add the monitor-array.sh

Array `tasks` are created on the fly when the `job` is monitored. This is done by a dedicated `monitor-array.sh` script. This script is broadly based on `monitor.sh`.

### DIFF
--- a/etc/job.yaml
+++ b/etc/job.yaml
@@ -86,18 +86,21 @@
 # submit_script_path: libexec/job/<scheduler>/submit.sh
 
 # ==============================================================================
-# Monitor Script Path
-# Specify the path to the script which monitors jobs.
+# Monitor Script Paths
+# Specify the paths to the scripts which monitor jobs.  Separate scripts are
+# used for monitoring array and non-array jobs.
 #
-# The default value will automatically use the 'scheduler' config. This feature
+# The default values will automatically use the 'scheduler' config. This feature
 # is not supported when overridden.
 #
 # Relative paths are expanded according to the "Relative path expansion" rules
 # described below.
 #
-# The environment variable flight_JOB_monitor_script_path takes precedence.
+# The environment variables flight_JOB_monitor_script_path and
+# flight_JOB_monitor_array_script_path take precedence.
 # ==============================================================================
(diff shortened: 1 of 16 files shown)


benarmston:
I've rebased on latest `master`.  I've added commits to fix getting the results dir for array and non-array jobs.  And to report errors when trying to view stdout/stderr for array jobs.  LGTM.

PR #44, opened 2021-09-03
Add a migration for the job metadata file

WilliamMcCumstie:
Based on #43 

---

This PR adds a migration script, which upgrade the `job` metadata from version 0 to version 1. This migration can be triggered in two ways:

* By running `run-migration`, or
* Implicitly when `job.valid?(:load)` runs.

---

## Dedicated Schema Definitions

The migrations is done by a `FlightJobMigration::Jobs` module. This has been deliberately decoupled from the `FlightJob::Job` class.

The migrations are intended to be a semi-permanent fixture for the current major release of `flight-job`. The `FlightJob::Job` should be able to change without affecting the migration code base.

To achieve this, the `FlightJob::Job::SCHEMAS` have been extracted into `config/schemas/job/version1.yaml`. This allows both `FlightJob` and `FlightJobMigration` to access the schema, without coupling to each other.

_NOTE:_ A YAML template is used to write the schema which is manually rendered to JSON. This is purely for development convince. The YAML isn't used by the code base.

---

## Version 0 Schemas

The old "version 0" `metadata` and `initial-metadata` schemas have been ported. When the migration is runs, the original data is verified against the appropriate schema.

This achieves two goals:

* Acts as a reference for what the migration needs to achieve, and
* Improves the logging for when a migration fails.

---

## Migrating SINGLETON or FAILED_SUBMISSION

Most jobs will be migrated to either the `SINGLETON` or `FAILED_SUBMISSION` job type. The migration basis this decision on the `exit_status`.

Most of this is fairly straight forward. The key change is `rendered_path` is now required. A dummy blank `rendered_path` is generated for legacy `job`.

Other changes:

* `*_time` are migrated based on the `state` of the job,
* `created_at` will default to epoch if unset,
* `std*_path` can not be empty string (but they can be `nil`), and
* `scheduler_id`/`scheduler_state` will both default to `unknown`.

A copy of the original metadata is maintained as `.metadata.v0.yaml` if the migration is successful.

These migrations use `File.write` as a race-condition is unlikely to cause data loss. The `run-monitor` script will be ran by `flight-job` if required.

---

## Migrating INITIALIZING jobs

There is a race condition if the job is migrated before the `metadata` has been generated. In these cases, the `initial-metadata` needs to be migrated.

This migration can cause data loss if `File.write` was used. Mainly the `submit.sh` script reports back in a separate process with the `scheduler_id`.

Instead, the a temporary file is `mv` into positioned. This is done without the force flag, and will not overwrite any existing data.


benarmston:
I've rebased on `master`.  I've added a commit extracting the validation for `Job` to `Job::Validator`.  This removes some duplicated validation code, groups the validation code together and reduces the amount of code `Job`.

LGTM.

PR #45, opened 2021-09-06
Implement the cancel-job command

WilliamMcCumstie:
A new `cancel-job` command has been added which wraps a `cancel_job_script`. For slurm, this directly calls `scontrol`. The monitor script is automatically called after a cancellation, which keeps things simple.

A prompt has been added to `delete-job` about the cancel command.


benarmston:
I've rebased on `master` and added support for cancelling array jobs.  LGTM.

PR #48, opened 2021-09-10
Allow results files to be followed

WilliamMcCumstie:
This PR adds three new flags to the `view-job-{results|stdout|stderr}` commands:

* `--follow`: Appends the data to the pager in real time
* `--retry`: Waits until the file exists, and
* `-F`: Equivalent to `--follow  --retry`

`FlightJob` does not implement the following feature itself. Instead it leverages `less +F`. This does mean the `PAGER` env var is ignored when following.

*NOTE:* An implementation where `FlightJob` writes to the `STDIN` of the pager was attempted, however this lead to a bad UX due to `less` caching the file.

---

Other changes:

* `TTY:Pager` has been removed in favour of a custom implementation, and
* `INTERRUPT` is ignored whilst the pager is open.


benarmston:
I've rebased on `master`; removed a `require 'tty-pager'` that prevented flight job from running; fixed a `pastel` issue that prevented waiting for a file.

benarmston:
I fixed a couple of issue that prevented this code from running.

 * The CLI was completely broken due to a `require 'tty-pager'` line being left in.
 * Paging was completely broken due to a `NameError: undefined local variable or method `pastel'` in Pager.

I've pushed commits for both of those.  Please make sure the code runs before creating a PR.

Also the `Pager#page` method contains a reference to `str` which doesn't exist.  I'm not sure what that branch is supposed to be doing.  Please take a look.

---

The `-F` option (not `+F`) doesn't really make sense when following the file.  The `-F` option is to quit if the file can be displayed on one screen.  When following a file we don't now if it will fit on a single screen, as we're assuming that it will grow.  Perhaps to a larger than one screen amount, perhaps not.

The `-F` option results in an awkward interaction if the user aborts out of following the file.  If the content currently, paged is less than a single screen `less` will exit, if more than a single screen `less` doesn't exit.  This is somewhat inconsistent and potentially surprising behaviour.  It also prevents the user from "pausing" the following by pressing `Ctrl-C` and then later "resuming" the following by pressing `F`.

This odd interaction had me really confused as to why `less` kept on aborting when I pressed `Ctrl-C`.

Also, the `-X` option is ignored when used with `+F`.  It should be removed.

WilliamMcCumstie:
> I've rebased on master; removed a require 'tty-pager' that prevented flight job from running; fixed a pastel issue that prevented waiting for a file

Apologise for that, slightly dodgy Friday afternoon commit.

In regards to testing, I discovered that `less +F` had some vary awkward interactions with `ctrl-c` when opened via `TTY::Pager`. I only discovered this after creating the PR. In interests of brevity, I didn't recreate the PR. However obviously some issues slipped by when removing `TTY::Pager`.

WilliamMcCumstie:
> Also the Pager#page method contains a reference to str which doesn't exist. I'm not sure what that branch is supposed to be doing. Please take a look.

Fixed so the pager supports regular strings:

https://github.com/openflighthpc/flight-job/pull/48/commits/3c161aa470899fe02148f0abbbf0d1cc2056639b

WilliamMcCumstie:
The `-FX` has been removed when following:

https://github.com/openflighthpc/flight-job/pull/48/commits/f49ce800817cbc55deac60f20fa4158985d0a48a

lib/flight_job/command.rb
```diff
@@ -73,10 +75,59 @@ def pastel
       @pastel ||= Pastel.new
     end
 
+    Pager = Struct.new(:retry_file, :follow, :pastel) do
+      def page(arg)
+        path = arg.is_a?(Hash) ? arg[:path] : arg.to_s
+        if path
+          open_path(path) { |io| page_io(io) }
+        else
+          io = StringIO.new(arg.to_s)
```
benarmston, line 84: This still doesn't make any sense to me.  As far as I can see `arg` can be one of three things

1. a Hash with a `:path` key with a truthy value
2. a Hash without a `:path` key or with a `:path` key with a falsy value
3. not a Hash

Only for case (2) do we take the `StringIO` route.  But I'm not sure why we're calling `to_s` on a Hash.

Am I missing something here?

benarmston (changes requested):

lib/flight_job/command.rb
WilliamMcCumstie: I agree the method signature is odd, but the `page` method is a drop in replacement for `TTY::Pager`. It needs to support two use cases:

* `pager.page(file: ...)`
* `pager.page("string")`

In general I dislike this type of signature as it is awkward to implement. However I want to avoid differing from `TTY:Pager` too much, as this will make retro-fitting other apps easier (if an when required).

---

As far as the `edge` case where `arg` is a `Hash` but `path` is falsey, _do not do it_. This doesn't appear in the code base and isn't supported.

If you prefer, I can modify the method signature to be:

```
def page(str = nil, path: nil)
end
```

However this uses an optional positional argument, which I also dislike.

WilliamMcCumstie: Updated the method signature, it should be _vaguely_ sane now:

679ef321242b9defc8e3b2b5cf0e161e42f12825

benarmston:
LGTM.

PR #49, opened 2021-09-13
Standardise the error message for missing results_dir

WilliamMcCumstie:
The distinction between a job not reporting its `results_dir` and the directory not existing as been removed.

Moving forward, all successfully submitted jobs will have a `results_dir`. A generic "no results found" is emitted when the directory does not exist. The same error is emitted for failed submissions.

### DIFF
--- a/lib/flight_job/command.rb
+++ b/lib/flight_job/command.rb
@@ -206,9 +206,9 @@ module FlightJob
       if !(job.results_dir && Dir.exists?(job.results_dir))
         case job.state
         when 'PENDING'
-          raise MissingError, 'Your job has not started, please try again latter...'
+          raise MissingError, 'Your job has not started, please try again later...'
         when *Job::RUNNING_STATES
-          raise MissingError, 'No job results found, please try again latter...'
+          raise MissingError, 'No job results found, please try again later...'
         else
           raise MissingError, 'No job results found.'
         end
--- a/lib/flight_job/command.rb
+++ b/lib/flight_job/command.rb
@@ -200,18 +200,27 @@ module FlightJob
       end
     end
 
-    def assert_results_dir_exists(job)
-      # NOTE: Jobs created with old versions of flight-job will not have a
-      # results directory.
-      if !(job.results_dir && Dir.exists?(job.results_dir))
-        case job.state
-        when 'PENDING'
-          raise MissingError, 'Your job has not started, please try again later...'
-        when *Job::RUNNING_STATES
-          raise MissingError, 'No job results found, please try again later...'
-        else
-          raise MissingError, 'No job results found.'
-        end
+    def assert_results_dir_exists(job, allow_empty: true)
+      error = if !job.results_dir
+                true
+              elsif !Dir.exists?(job.results_dir)
+                true
+              elsif allow_empty
+                false
+              elsif Dir.empty?(job.results_dir)
+                true
+              else
+                false
+              end
+      return unless error
+
+      case job.state
+      when 'PENDING'
+        raise MissingError, 'Your job has not started, please try again later...'
+      when *Job::RUNNING_STATES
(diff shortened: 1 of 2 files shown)


benarmston:
Rebased on `master`.

lib/flight_job/commands/list_job_results.rb
```diff
@@ -37,7 +37,13 @@ class ListJobResults < Command
 
       def run
         job = load_job(args.first)
-        assert_results_dir_exists(job)
+
+        # Skip the assertion check when hard wrapping and the directory exists
+        # This causes to command to run on empty directories
+        if hard_wrap? && Dir.exists?(job.results_dir)
```
benarmston, line 43: I don't think this is doing what you want it to.

benarmston: Also `hard_wrap` is a confusing name.  I'm not sure what part of passing additional options to the `ls` command is hard or why its wrapped.  Or is hard wrap the opposite of a soft wrap?

Can we get a better name for this please?

Also let's rename `HARD_ERROR` to `LS_ERROR`.

lib/flight_job/command.rb
```diff
@@ -203,8 +203,15 @@ def load_job(id)
     def assert_results_dir_exists(job)
       # NOTE: Jobs created with old versions of flight-job will not have a
       # results directory.
-      unless job.results_dir
-        raise MissingError, "The job did not report its results directory"
+      if !(job.results_dir && Dir.exists?(job.results_dir))
+        case job.state
+        when 'PENDING'
+          raise MissingError, 'Your job has not started, please try again latter...'
```
benarmston, line 209: `latter` should be spelled `later`.  Here and elsewhere please.

lib/flight_job/command.rb
```diff
@@ -203,8 +203,15 @@ def load_job(id)
...
+        when *Job::RUNNING_STATES
+          raise MissingError, 'No job results found, please try again latter...'
+        else
+          raise MissingError, 'No job results found.'
+        end
```
benarmston, line 214: These error messages can also be printed directly by the `ListJobResultsCommand#run` method.  When printed by the `run` method they are coloured.  When raised here they aren't.  Can we get these coloured too please?

lib/flight_job/commands/list_job_results.rb
benarmston: Also, why do we want this branch.  `assert_results_dir_exists` doesn't check for empty directories.

benarmston (changes requested):
This doesn't work.  I still get the raw `ls` error aboiut the directory not existing.

lib/flight_job/command.rb
WilliamMcCumstie: 57a3f08de3ee67b0da9e0992f505c22fad679b9b

lib/flight_job/commands/list_job_results.rb
WilliamMcCumstie: I've added the assertion check back here: 57a3f08de3ee67b0da9e0992f505c22fad679b9b.

I originally removed the check when hard wrapping `ls` to preserve the original error. However this isn't required.

WilliamMcCumstie: Renamed `LS_ERROR`: 583e9de2742b3684a28248ece69a625c1ab8674f

lib/flight_job/command.rb
WilliamMcCumstie: I am not entirely sure why the colourisation isn't working. The top level error handler should be applying the default colours. This _may_ be an issue with your `TERM` setting as I can not reproduce it.

---


Regardless, I have collapsed the two sets of warnings/errors into a single method:
7f2ce35fef7732fdf6542cf7f9233dbfdf5113af

benarmston:
LGTM.

PR #51, opened 2021-09-17
Wip/extract template spec

WilliamMcCumstie:
Based on #46 

---

The specs are getting quite long for the `template` metadata. This PR is migrating it to yaml file

This should make following the spec a bit easier


benarmston:
Closing this as it has largely been replaced with recent refactorings of template.

PR #54, opened 2021-09-30
Refactor the transition classes

WilliamMcCumstie:
This PR reworks the transition classes to be a bit cleaner:

* Renames the classes to be based on nouns
* Removes the use of `SimpleDelegator`
* Error on `run!` if the transition fails, and
* Allow errors in the `run` methods.

The `Job#monitor` method now calls the `run` version of the transitions. The monitor can always be re-ran at a future date, so errors are allowed. The metadata is however reset if the transition fails.

The `cancel`/`submit` methods call the `run!` to trigger a command error.

---

Also fixed a bug where terminal array jobs where always monitored.

### DIFF
--- a/lib/flight_job/job_transitions/array_monitor.rb
+++ b/lib/flight_job/job_transitions/array_monitor.rb
@@ -62,6 +62,11 @@ module FlightJob
       end
 
       def run!
+        if job.terminal?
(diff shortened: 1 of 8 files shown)


WilliamMcCumstie:
Rebased on master

benarmston:
LGTM.

PR #56, opened 2021-09-30
Save the jobs after they have been submitted

WilliamMcCumstie:
Previously a job had to be submitted and monitored before being saved. This meant `flight-job` could loose track of submitted jobs.

This fixes the issue by:

* Renames the original `INITIALIZING` to submitting, and
* Saves successfully submitted jobs with the `MONITORING` job type.

A `MONITORING` job means it was successfully submitted but not yet been monitored for the first time. They are hidden from the end user (similar to `SUBMITTING`) to prevent issues with null fields.

### DIFF
--- a/config/schemas/job/version1.json
+++ b/config/schemas/job/version1.json
@@ -27,6 +27,7 @@
     "job_type": {
       "enum": [
         "INITIALIZING",
+        "MONITORING",
         "SINGLETON",
         "ARRAY",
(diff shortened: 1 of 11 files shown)


WilliamMcCumstie:
Rebased on #54

benarmston:
Fixed a couple of issues, and renamed `MONITORING` -> `BOOTSTRAPPING`.

PR #57, opened 2021-10-01
Add the cancelling job state

WilliamMcCumstie:
Based on #56 

Previously a job would report as `RUNNING` after `cancel-job` was called. This is because the slurm `COMPLETING` state was being mapped to `RUNNING`. Solving this issue is tricky as all jobs pass through the slurm `COMPLETING` state regardless if they have been cancelled.

---

The `monitor.sh` and `monitor-array.sh` scripts can now report back jobs with a `COMPLETING` state. This is an internal state which is hidden from the end user. It will be reported back as either `RUNNING` or `CANCELLING`.

When `cancel-job` runs, it sets a `cancelling` flag on the `job` (both singleton and array). This is used to toggle the `COMPLETING` state between `RUNNING`/`CANCELLING`. It does not affect any other states.

---

*NOTE*:
The legacy `cancelled` flag on array jobs is now obsolete. It has been deprecated and is otherwise ignored.

It has not been removed as this would be *technically* a breaking change (despite the work being otherwise unreleased).


WilliamMcCumstie:
Rebased on #56

benarmston:
Rebased on `master`.

benarmston:
LGTM.

PR #58, opened 2021-10-01
Run both `sacct` and `scontrol` for array jobs

WilliamMcCumstie:
Based on #57

For sufficiently long running array jobs, the first task(s) may drop out of `scontrol` before the last tasks finishes. This can cause jobs to get stuck in a `RUNNING` state as the initial tasks cannot be updated. This will rectify itself when the entire job moves to `sacct`.

Instead, the `monitor-array.sh` runs both `scontrol` and `sacct` and merges the tasks list.

### DIFF
--- a/libexec/job/slurm/monitor-array.sh
+++ b/libexec/job/slurm/monitor-array.sh
@@ -114,7 +114,7 @@ run_scontrol() {
 }
 
 run_sacct() {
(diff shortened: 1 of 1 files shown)


WilliamMcCumstie:
Rebased on #57

benarmston:
I've rebased on `master` and fixed an issue in which array jobs were not being correctly identified as completed.

PR #62, opened 2021-10-05
Condense the `list-job` output and upgrade `OutputMode`

WilliamMcCumstie:
This PR does two things:

* Upgrade `OutputMode` to use the new `Formatter` interface, and
* Reduce the table size of the interactive `list-jobs` output

---

## OuputMode::Formatter interface

This is a new interface that provides two major improvements:

* The flag methods: `interactive?`, `verbose?`, `ascii?`, and `color?`; and
* Will apply the `yes`, `no`, `default`, and `time` formatters.

The older style of specifying the flag methods inline is no longer supported. Instead standard branching logic should be used:

```
# Not Supported
  register_attribute(header: ..., verbose: true) { ... }

# New Style
  if verbose?
    register(header: ...) { ... }
  end
```

The `register` method adds a `block` as a column/attribute to the output. It will automatically wrap the block with the `format` method; which applies the `yes`/`no`/`default`/`time` formatters. Custom formatters can be implemented by overriding the method.

*NOTE:* The `constructor` block is a shorthand to `instance_exec` making the following equivalent:

```
# Using new
class Output < OutputMode::Fomatter; end
output = Output.new
output.register
...
output.render

# Using the constructor
class Output
  constructor do
    register
    ...
  end
end
Output.render
```

---

## Condense the list-jobs output

The `list-jobs` interactive outputs have been modified to reduce their size. In particular:

* The non-interactive output has not been modified,
* Submit status has been dropped from interactive outputs,
* The actual/estimated times share a column in the simplified interactive output, and
* _date-times_ are shortened to time only if they are on the same day, or date only for past/future dates (interactive only).


lib/flight_job/outputs/info_job.rb
```diff
@@ -28,107 +28,99 @@
...
-    register_attribute(header: 'Submit Status', verbose: true) { |j| j.submit_status.to_s }
-
-    register_attribute(header: 'Submitted at') do |job, verbose:|
-      if verbose
-        job.created_at
-      else
-        DateTime.rfc3339(job.created_at).strftime('%d/%m/%y %H:%M')
-      end
+  class Outputs::InfoJob < OutputMode::Formatters::Show
+    def initialize(job, submit: true, **opts)
```
benarmston, line 32: The stdout and stderr is being included always.  I suspect its due to `submit: true` being the default here.

lib/flight_job/outputs/info_job.rb
```diff
@@ -28,107 +28,99 @@
...
+
+    constructor do
+      template(<<~ERB) if interactive?
+        <% each(:default) do |value, padding:, field:| -%>
+        <%= format_row(field, value, padding) %>
+        <% end -%>
+        <%
+          submit_outputs = callables.select { |proc| proc.config[:section] == :submit }
+          if verbose? || submit?
+        -%>
```
benarmston, line 58: What's the reason for special handling of the stdout and stderr in the template?  Wouldn't it be easier to branch in the `constructor` block?

```ruby
constructor do
  # ....
  if verbose? || submit?
    register(header: 'Submit stdout') { job.submit_stdout }
  end
end
```

lib/flight_job/outputs/info_job.rb
```diff
@@ -28,107 +28,99 @@
...
+      register(header: 'Results Dir') { job.results_dir }
+      if verbose?
+        register(header: 'Estimated start', &:estimated_start_time)
+        register(header: 'Estimated end', &:estimated_end_time)
+      end
 
-    def self.build_output(**opts)
-      submit = opts.delete(:submit)
-      super(template: TEMPLATE, context: { submit: submit }, **opts)
+      register(header: "Desktop ID:", &:desktop_id)
```
benarmston, line 123: The `:` needs removing.

lib/flight_job/outputs/info_job.rb
```diff
@@ -28,107 +28,99 @@
...
+      register(header: 'State') { job.state }
+
+      # Show a boolean in the "simplified" output, and the exit code in the verbose
+      # NOTE: There is a rendering issue of integers into the TSV output. Needs investigation
+      if verbose?
+        register(header: 'Submit Status') { job.submit_status.to_s }
       else
-        Outputs.format_time(job.estimated_start_time, false)
+        register(header: 'Submitted') { job.submit_status == 0 }
       end
```
benarmston, line 78: It's really hard to follow what's happening form here on down.  There are 5 different conditionals branched on for a total of 6 branches.  Something like the following would be much clearer:

```ruby
constructor do
  if verbose?
    register_common
    # ... register other cols
  elsif interactive?
    register_common
    # ... register other cols
  else
    register_common
    # ... register other cols
  end
end
```

lib/flight_job/outputs/list_array_tasks.rb
```diff
@@ -29,28 +29,27 @@
...
-
-    register_column(header: 'Started at') do |task, verbose:|
-      Outputs.format_time(task.metadata['start_time'], verbose)
-    end
-    register_column(header: 'Ended at') do |task, verbose:|
-      Outputs.format_time(task.metadata['end_time'], verbose)
+  class Outputs::ListArrayTasks < OutputMode::Formatters::Index
+    def fetch_time(task, key)
+      time = task.metadata[key]
+      time ? Time.parse(time) : time
```
benarmston, line 35: Same question here as before.

lib/flight_job/outputs/list_templates.rb
```diff
@@ -28,28 +28,27 @@
...
-    end
+      register(header: 'Name') do |template|
+        template.id
+      end
+      file_header = "File (Dir: #{FlightJob.config.templates_dir})"
 
-    def self.build_output(**opts)
-      super(row_color: :cyan, header_color: :bold, **opts)
+      if verbose?
+        register_column(header: file_header) do |template|
```
benarmston, line 44: This is broke.  Do you mean `register`?

lib/flight_job/outputs/list_templates.rb
```diff
@@ -28,28 +28,27 @@
...
-    register_column(header: file_header, verbose: true) do |template, interactive:|
-      if interactive
-        Pathname.new(template.workload_path).relative_path_from FlightJob.config.templates_dir
-      else
-        template.workload_path
+  class Outputs::ListTemplates < OutputMode::Formatters::Index
+    constructor do
+      register(header: 'Index', row_color: :yellow) do |template|
+        # NOTE: The OutputMode library does not support *_with_index type notation
+        #       Instead the index needs to be cached on the object itself
```
benarmston, line 35: What does this comment mean?  Why is it relevant here?  Everywhere else seems to have blocks such as `{ |object| object.foo }`.  Why does that need commenting on in this case?

lib/flight_job/outputs/info_job.rb
WilliamMcCumstie: The formatting is slightly different. The other keys are rendered inline: `ID: foo-job`

Where the `STDOUT`/`STDERR` are offline:

```
STDOUT:
<content>
```

---

There needs to be special casing regardless, so I embedded it into the template

lib/flight_job/outputs/info_job.rb
WilliamMcCumstie: Will look what I can do, however at best it will end up:

```
if verbose?
  ...
elsif interactive?
  ...
  job.actual_start_time ? register_actual_* : register_estimated_*
  job.actual_end_time ? register_actual_* : register_estimated_*
else
  ...
end
```

lib/flight_job/outputs/info_job.rb
WilliamMcCumstie: Cleaned the template up a bit:

9419cb7ac23e717c30620de954207dbd5df3e954

lib/flight_job/outputs/info_job.rb
WilliamMcCumstie: 4a6fc11c3917734bedb2b2088ae338a39fbdfa68

lib/flight_job/outputs/list_array_tasks.rb
WilliamMcCumstie: Ditto

lib/flight_job/outputs/list_templates.rb
WilliamMcCumstie: 73388a26fc3dafb4d3666b87ba91a014ef43768c

lib/flight_job/outputs/list_templates.rb
WilliamMcCumstie: Yes that is correct, removed the comment

154ccf504649a3702ef33fcd7f85bf867ed7e0ec

WilliamMcCumstie:
In regard to your block comments, two more changes have been made:

* The `interactive?` flag has been renamed `humanize?`, and
* The `constructor do` mechanism has been replaced with `register_all`.

PR #68, opened 2021-11-02
Don't blank environment in interactive sessions

VoxSecundus:
This PR amends the `flight desktop start` call when starting an interactive session, calling a CLI parameter introduced to `desktop` in [this PR](https://github.com/openflighthpc/flight-desktop/pull/40) which prevents the session environment from being blanked, and allows the host environment to persist.


benarmston:
LGTM.

PR #71, opened 2021-11-15
Add support for configurable results directory

VoxSecundus:
This PR contains initial work to complete the work required for [adding support for a configurable results directory to Flight Job.](https://trello.com/c/IpNzQzWn/388-add-support-for-configurable-results-directory-to-flight-job). The work specified in Phase 4 of the card is yet to be completed; the rest is implemented here.

### DIFF
--- a/lib/flight_job/models/job.rb
+++ b/lib/flight_job/models/job.rb
@@ -179,7 +179,7 @@ module FlightJob
     end
 
     def results_dir
-      self.controls_file("results_dir").read || metadata['results_dir']
+      controls_file("results_dir").read || metadata['results_dir']
     end
 
     # DEPRECATED: This method belongs on the decorator!
--- a/lib/flight_job/models/job.rb
+++ b/lib/flight_job/models/job.rb
@@ -179,7 +179,7 @@ module FlightJob
     end
 
     def results_dir
-      metadata['results_dir']
+      self.controls_file("results_dir") || metadata['results_dir']
     end
 
     # DEPRECATED: This method belongs on the decorator!
--- a/lib/flight_job/models/job.rb
+++ b/lib/flight_job/models/job.rb
@@ -179,7 +179,7 @@ module FlightJob
     end
 
     def results_dir
-      self.controls_file("results_dir") || metadata['results_dir']
+      self.controls_file("results_dir").read || metadata['results_dir']
     end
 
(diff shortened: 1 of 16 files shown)


usr/share/job/adapter.slurm.erb
```diff
@@ -56,7 +52,8 @@ session_start() {
     echo "Starting session..."
     (
         set -o pipefail
-        ${flight_ROOT:-/opt/flight}/bin/flight desktop \
+        #${flight_ROOT:-/opt/flight}/bin/flight desktop \
+        /opt/flight/bin/ruby /code/flight-desktop/bin/desktop \
```
benarmston, line 56: Accidental commit alert!

usr/share/job/templates/simple-array/workload.erb
```diff
@@ -38,6 +38,8 @@ source "${flight_ROOT:-/opt/flight}"/etc/setup.sh
 # will be unable to help you access your results.
 #
 # RESULTS_DIR has already been set to "$WORKING_DIR/${SLURM_JOB_NAME}-outputs/$SLURM_ARRAY_JOB_ID"
+RESULTS_DIR="$(pwd)/${SLURM_JOB-NAME}-outputs/$SLURM_ARRAY_JOB_ID"
```
benarmston, line 41: Can we remove the line claiming `RESULTS_DIR` has already been set.

lib/flight_job/models/job.rb
```diff
@@ -179,7 +179,7 @@ def created_at
     end
 
     def results_dir
-      metadata['results_dir']
+      self.controls_file("results_dir").read || metadata['results_dir']
```
benarmston, line 182: We typically don't add `self.` to method calls unless there is a good reason to do so.  I don't see any reason why it needs to be here.  Can you remove it please.

benarmston (changes requested):
This looks to be headed in the right direction.  Can you take a look at the comments that I've added.  Thanks.

benarmston (changes requested):
All of our templates contain the warning `(or if RESULTS_DIR is modified)` just above the line setting `RESULTS_DIR`.  That warning used to be true, but it no longer is.  Please remove it from all templates.

benarmston:
LGTM.  Thanks @VoxSecundus

PR #75, opened 2021-11-29
Warn user when copying a template without defaults

VoxSecundus:
This PR aims to add a warning given to the user when copying a template containing questions without a default value. When copying such a template, `stderr` is given a warning and a list of questions which are missing a default value. The copy takes place nonetheless.

The PR also fixes a related issue where a script could not be rendered if it did not have an ID (which it does not when rendering for the purpose of writing to a new template file).

### DIFF
--- a/lib/flight_job/commands/copy_template.rb
+++ b/lib/flight_job/commands/copy_template.rb
@@ -29,6 +29,14 @@ module FlightJob
   module Commands
     class CopyTemplate < Command
       def run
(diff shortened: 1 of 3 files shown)


benarmston (changes requested):
With the exception of the below comment this looks good to me.

benarmston:
The warning is now printed in yellow.  Otherwise this LGTM.

PR #76, opened 2021-12-02
Mark invalid templates when listing templates

VoxSecundus:
This PR introduces a pair of changes to the template validation workflow in Flight Job.

The first of these changes removes the separation of 'verbose' and 'regular' validations. All `.valid?` calls on a `template` run the same set of validations every time (which is now all of the previously separate validations grouped into one).

The second change adds validations to the `list-templates` command. Now, when listing templates, if a template is invalid, its row in the resulting table will be highlighted and marked as being invalid.


benarmston:
This looks good to me.  Thanks for your work on this @VoxSecundus

PR #77, opened 2021-12-06
Re-implement old method of grabbing results directory

VoxSecundus:
Backwards compatibility dictates that the new version of the application must be able to accommodate scripts created via the old version. Newly created scripts should now have an automatically generated results directory in the job metadata, as well as in the controls directory. The controls directory method is always preferred; however, when the results directory control file is not present (as is the case for version 1 scripts), the application falls back to the results directory present in the job metadata.

### DIFF
--- a/config/schemas/job/version2.json
+++ b/config/schemas/job/version2.json
@@ -243,6 +243,10 @@
           "type": "string",
           "minLength": 1
         },
+        "results_dir": {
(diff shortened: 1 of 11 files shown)


benarmston:
I've added a commit fixing a couple of issues with checking and setting controls dir.  This now LGTM.

PR #85, opened 2022-04-01
Info templates change

megan-godfrey:
Now gives output similar to info-script. License and copyright only shown if verbose option active

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ vendor
 .env.development.local
 /log
 tags
+*.swp
--- a/lib/flight_job/models/template.rb
+++ b/lib/flight_job/models/template.rb
@@ -226,6 +226,18 @@ module FlightJob
        metadata['name']
     end
 
+    def copyright
+       metadata['copyright']
+    end
+    
+    def license
+       metadata['license']
+    end
+    
+    def description
+       metadata['description']
+    end
+    
     def tags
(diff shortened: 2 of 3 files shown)


benarmston (changes requested):
A file at `lib/flight_job/models/.template.rb.swp` has been committed by mistake.  This is a file that vim creates FSR whilst editing a file.  We should have a rule in `.gitignore` that prevents such files from being committed.

Can you delete this file (bonus points for rebasing so it appears to never have been committed) and add a rule to `.gitignore` in the top level directory to ignore `*.swp` files.

benarmston:
A couple of things to look at here.  Rebasing to avoid having that first "broken" commit and removal of the `.swp` file.  Other than that this LGTM.

benarmston:
Thanks for rebasing this @megan-godfrey .  I've added the change to the `.gitignore` which you can see at https://github.com/openflighthpc/flight-job/pull/85/commits/5a9cb7a24f00f021735700118a8cdbc628a7ab36

This LGTM.

benarmston:
Thanks for rebasing this @megan-godfrey .  I've added the change to the `.gitignore` which you can see at https://github.com/openflighthpc/flight-job/pull/85/commits/5a9cb7a24f00f021735700118a8cdbc628a7ab36

This LGTM.

PR #87, opened 2022-04-04
Markdown descriptions

megan-godfrey:
For templates, added the ability to read description_markdown field from metadata.yaml, if it exists. The description will then be rendered as markdown if the humanize? option is active. If the description_markdown field doesn't exist, the plain text version is read in as before. I added an example description_markdown field to the simple-array template.

For scripts, no separate notes file dedicated to markdown, just rendering as markdown if the humanize option is active. I didn't think it made sense to have separate versions here for markdown/ plain text, but let me know if you want me to add that.

### DIFF
--- a/lib/flight_job/outputs/info_script.rb
+++ b/lib/flight_job/outputs/info_script.rb
@@ -26,9 +26,11 @@
 #==============================================================================
 
 require 'output_mode'
+require_relative '../markdown_renderer'
 
 module FlightJob
   class Outputs::InfoScript < OutputMode::Formatters::Show
+
     def register_all
       template(<<~ERB) if humanize?
         <% each(:main) do |value, field:, padding:, **_| -%>
@@ -51,7 +53,13 @@ module FlightJob
         Time.parse script.created_at
       end
 
-      register(section: :notes, header: 'Notes') { |s| s.notes }
+      if humanize?
+        notes = MarkdownRenderer.new(object.notes).wrap_markdown
+      else
+        notes = object.notes
+      end
+      register(section: :notes, header: 'Notes') { notes }
     end
   end
 end
+
--- a/lib/flight_job/models/template.rb
+++ b/lib/flight_job/models/template.rb
@@ -47,6 +47,7 @@ module FlightJob
       "properties" => {
         "copyright" => { "type" => "string" },
         "description" => { "type" => "string" },
+        "description_markdown" => { "type" => "string" },
         "generation_questions" => {
           "type" => "array",
           "items" => { "$ref" => "#/$defs/question_def" }
@@ -236,6 +237,10 @@ module FlightJob
     
     def description
        metadata['description']
+    end
+
+    def description_markdown
+      metadata['description_markdown'] ||= metadata['description']
     end
     
     def tags
--- a/lib/flight_job/models/template.rb
+++ b/lib/flight_job/models/template.rb
@@ -234,6 +234,10 @@ module FlightJob
        metadata['license']
     end
     
+    def synopsis
+       metadata['synopsis']
+    end
+
     def description
        metadata['description']
     end
--- a/lib/flight_job/models/template.rb
+++ b/lib/flight_job/models/template.rb
@@ -47,7 +47,6 @@ module FlightJob
       "properties" => {
         "copyright" => { "type" => "string" },
         "description" => { "type" => "string" },
-        "description_markdown" => { "type" => "string" },
         "generation_questions" => {
           "type" => "array",
           "items" => { "$ref" => "#/$defs/question_def" }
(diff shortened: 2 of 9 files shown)


lib/flight_job/outputs/info_script.rb
```diff
@@ -51,7 +53,13 @@ def register_all
         Time.parse script.created_at
       end
 
-      register(section: :notes, header: 'Notes') { |s| s.notes }
+      if humanize?
+        notes = MarkdownRenderer.new(object.notes).wrap_markdown
+      else
+        notes = object.notes
+      end
+      register(section: :notes, header: 'Notes') { notes }
```
benarmston, line 61: It's worth noting that this method of first branching to determine how the notes should look and *then* registering the column works fine when a single object is being displayed such as an `info-*` command.  However, this would not work for a `list-*` command which is rendering multiple objects.

The code is fine as it is, but its probably worth making a mental note of the following.

For a `list-*` command the following would be needed:

```ruby
register(section: :notes, header: 'Notes') do |s|
  if humanize?
    MarkdownRenderer.new(s.notes).wrap_markdown
  else
    s.notes
  end
end
```

The block provided to `register` is called once for each item in the list and each item (with the above) would use its own notes.  With the way that you have it, the same set of notes would be used for each item in the list.

(The above will also work just fine for an `info-*` command).

benarmston, line 61: It's worth noting that this method of first branching to determine how the notes should look and *then* registering the column works fine when a single object is being displayed such as an `info-*` command.  However, this would not work for a `list-*` command which is rendering multiple objects.

The code is fine as it is, but its probably worth making a mental note of the following.

For a `list-*` command the following would be needed:

```ruby
register(section: :notes, header: 'Notes') do |s|
  if humanize?
    MarkdownRenderer.new(s.notes).wrap_markdown
  else
    s.notes
  end
end
```

The block provided to `register` is called once for each item in the list and each item (with the above) would use its own notes.  With the way that you have it, the same set of notes would be used for each item in the list.

(The above will also work just fine for an `info-*` command).

benarmston (changes requested):
Thanks for your work on this. You've mostly implemented what was wanted, but have also created a `description_markdown` attribute on the template that isn't needed.  Perhaps there is some confusion on what Markdown is or what was wanted here.

---

First, let's have a bit of an aside on what markdown is.

Markdown is a light-weight way of marking up how text should be laid out.  It can be compiled to a number of formats including HTML, but also it can be compiled to "shell escape codes" (properly called ANSII escape codes) which can be used to make terminal output pretty.

For instance, the following document:

```markdown
# A header

A paragraph with some *emphasised* text, and a [link to google](www.google.com).
```

could be compiled to the following HTML:

```html
<h1>A header</h1>
<p>A paragraph with some <em>emphasised</em> text, and a <a href="www.google.com">link to google</a>.
```

or it could alternatively be compiled to the following set of ANSII escape
codes:

```sh
\e[36;1;4mA header\e[0m

A paragraph with some \e[33memphasised\e[0m text, and a link to google \u00BB \e[33;4mwww.google.com\e[0m.
```

Both the HTML and the ANSII escape code output are not Markdown.  Markdown is the source document that they are created from.  I think that you already knew most of that, but its useful to labour the point that HTML produced from Markdown isn't Markdown.

---

When the template is displayed in the webapp, the webapp takes the raw markdown and the webapp compiles it to HTML for a pretty display.  This is already handled by the webapp.

Currently, the CLI displays the raw markdown.  Displaying the raw markdown currently looks fine, because our template descriptions make little use of markdown.

However, if our templates were to make better use of Markdown, displaying the raw Markdown to users could look odd or even broken to some users.  So this work is a prerequisite to allowing the template descriptions to use more interesting Markdown.

I think that some of the confusion may have come about due to the confusion due to the limited use of Markdown in the templates.  So let's change the description for `mpi-nodes` to the following and run `info-template mpi-nodes` both with and without your change.

```yaml
description: |
  This may be because you intend to spawn multiple threads yourself or intend
  to run an [MPI](https://en.wikipedia.org/wiki/Message_Passing_Interface)
  application. This method is often used for MPI jobs where you need to
  specify the number of processes, machine file and process allocation
  manually, rather than leave this up to the scheduler.

  Your job will be allocated the requested number of *empty nodes*. This
  method should not be used for SMP (OpenMP) jobs. Only use this method if you
  need exclusive use of each node allocated; prefer a multiple slot job for a
  shorter queue time.
```

Before this PR

image

With this PR

image

---

All of that is a long winded way of saying, 

* the `description_markdown` field isn't needed on the template `metadata.yaml` file,
* the two references to `description_markdown` aren't needed in the `template.rb` file.
* `info_template.rb` should be using `description` instead of `description_markdown`.

Hopefully, that clarifies rather than confuses.

benarmston:
Thanks for your work on this @megan-godfrey .  I've added a couple of commits, which update the template descriptions and also include the synopsis in the output of `info-template`.  This LGTM.

PR #88, opened 2022-04-05
Default script notes

megan-godfrey:
Sets default notes for created scripts based on the template, if none are given using the --notes flag.

notes.md file created in each template directory (/code/flight-job/usr/share/job/templates/*/notes.md), except for the smp directory (left out for testing)


benarmston:
I've added a commit updating the default template notes and removing an accidentally committed file.  This LGTM now.

PR #89, opened 2022-04-06
Refactor job's metadata

benarmston:
This PR extracts metadata handling code from `Job` into `Job::Metadata`.

Previously, `Job` stored the result of its metadata in a `@metadata` instance variable as a `Hash`.

Now, the job's metadata is handled by a new `Job::Metadata` class.  This class is responsible for loading, validating, saving and providing access to its content.  It is also responsible for determining if the job's metadata has been persisted?

There are three class methods to load the metadata `load_from_path`, `blank` and `from_script`.  `load_from_path` is used for loading the metadata for existing jobs.  `from_script` is used for initializing a job's metadata from a script when it is first created.  `blank` is used to create an empty (or balnk) metadata instance.

The blank metadata is useful for checking if the metadata has been peristed.  Having a simple mechanism for this has enabled cleaning up the validations and making some branches on whether `@metadata` was set clearer as to their purpose.

`Job::Metadata` is explicit in what attributes it expects to be present in the metadata.  Access to the metadata's content is limited to the listed attributes.  `Job::Metadata` provides the same `[]`/`[]=` access as the previous hash, however, the keys are checked against the listed attributes.  This is intended to help catch typos such as `job.metadata['shceduler_id']` instead of `job.metadata['scheduler_id']`.

Previously, `Job` delegated validating the metadata to `Job::Validator` and made a distinction between load-time validations and save-time validations.  Now it is `Job::Metadata` that delegates to `Job::Validator` and there is no distinction between load-time and save-time validations.

The load/save validation distinction was previously used to allow running some callbacks when loading a job after it had been validated.  Some of them, such as adjusting the active index file were previously reworked to use an `after_initialize` callback.  This refactoring allows for the final behaviour that was (mis-)using validation callbacks, namely migrating metadata from one schema to another, to be migrated to an `after_initialize` callback.

`Job::Metadata` does have two flavours of validation.  The standard `valid?` flavour and a `valid?(:metadata_type)` validation that checks if the metadata is a `Hash`.  This is used in a single location and I can't think of a way to avoid it.

Much of the behaviour of `Job::Metadata` is derived from its base class `MetadataBase`.  The intention is that `Template`, `Script` and `Task` will later use a similar mechanism to handle their own metadata.

Some unused methods have also been removed from `Job`: `active_metadata` and `std{out,err}_path!`.


benarmston:
@VoxSecundus could you take a look at this PR and see if you can break anything please?

VoxSecundus:
@benarmston I haven't been able to break this thus far. The only comment I have is with regards to the expected attributes of a job's metadata; when a metadata file contains attributes that are not listed in the `attributes` call in `FlightJob::Job::Metadata`, it is ignored when running something like `info-job`. Is this expected, or should the 'explicitness' of the metadata's attributes include coverage for **un**expected attributes?

PR #91, opened 2022-04-22
Desktop names from job ID

megan-godfrey:
Automatically names interactive desktop sessions on login and compute nodes, where the name is the job ID.

I've left the path pointing to the dev version of flight desktop in `/code/flight-job/usr/share/job/adapter.slurm.erb` for testing, I wasn't sure when would be a good time to put it back to pointing to the rpm version.


megan-godfrey:
I'm sorry about all the old commits - I used `git rebase --interactive` to remove them on my machine, but now I've pushed to the online repo they're all there again :/ not sure what's happened?

PR #92, opened 2022-04-22
Enh/desktop names

megan-godfrey:
Automatically names interactive desktop sessions on login and compute nodes, where the name is the job ID.

I've left the path pointing to the dev version of flight desktop in /code/flight-job/usr/share/job/adapter.slurm.erb for testing, I wasn't sure when would be a good time to put it back to pointing to the rpm version.

### DIFF
--- a/lib/flight_job/desktop_cli.rb
+++ b/lib/flight_job/desktop_cli.rb
@@ -39,6 +39,7 @@ module FlightJob
           "--no-override-env",
           "--script", script,
           "--kill-on-script-exit",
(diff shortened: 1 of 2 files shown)


megan-godfrey:
Sorry I shall fix commits after lunch

benarmston:
I've added a commit reinstating usage of the RPM version of desktop. This is going to break newly created interactive job scripts until a new version of Flight Desktop is packaged and installed.  The workaround is to edit the job script (via `job edit-script SCRIPT_ID`)  and adjust the path to the desktop command accordingly.

PR #97, opened 2022-05-04
Wip/refactor metadata

megan-godfrey:


benarmston:
There are some `.idea/*` files that have ended up in the PR.  Before this PR is ready for review, you'll need to remove the files and update the `.gitignore`.

spec/models/script_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    %w(created_at script_name tags template_id).each do |attr|
+      it "reads #{attr} from the metadata" do
+        expect(script.send(attr)).to eq(metadata[attr])
+      end
+    end
+
+    it "reads submission_answers from the metadata defaulting to {}" do
+      expect(script.answers).to eq(metadata["answers"] || {})
+    end
```
benarmston, line 61: These answers are `generation_answers` not `submission_answers`.  Generation answers are the answers given when generating the script from the template.  Submission answers are the answers given when submitting the script to the scheduler.

spec/models/script_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+end
```
benarmston, line 63: Can you add specs to check that writing the metadata works.  That is write a value that is stored in `@metadata` and then check that it can be read back.  Something like:

```ruby
{
      template_id: 'foo',
    }.each do |attr, value|
      it "writes #{attr} to the metadata" do
        expect(script.send(attr)).not_to eq(value)
        script.send("#{attr}=", value)
        expect(script.send(attr)).to eq(value)
      end
    end
```

benarmston (commented):
Those specs look good.  They're missing specs to check that writing to `@metadata` works.

spec/models/job_spec.rb
```diff
@@ -13,6 +13,7 @@
 
       before(:each) { subject.valid? }
 
+      puts config.jobs_dir
```
benarmston, line 16: This line is causing the job specs to fail.  `config` is defined as a `let` block so there are some restrictions on when it can be called.  You can call it from inside an `it` block or inside a `before(:each)` block, but not directly inside a `context` block.

PR #99, opened 2022-05-09
All but one warning suppressed

megan-godfrey:


benarmston (changes requested):
Reading the spec output is so much better now.  Thanks.

I've got some comments on how the duplicated method errors have been suppressed.  Could you look at them.

Also, could you look at the warning `/code/flight-job/lib/flight_job/models/job/metadata.rb:93: warning: instance variable @save_point not initialized`.    That's generated by running the specs in `spec/models/job/metadata_spec.rb`.

There is a warning about `raise_error` not having been given a specific error.  Could you change the line `spec/models/job/metadata_spec.rb:106` to `expect { metadata.save }.to raise_error(FlightJob::InternalError)`.

benarmston:
The changes here look good.  When #98 is merged, I'll rebase this and see if all warning are addressed in the specs on that PR too.

benarmston:
I've rebased this on the latest `master` which has introduced a whole host of new warnings, as a result of more code being covered by the specs.

PR #100, opened 2022-05-13
Dev/display invalid jobs

megan-godfrey:
BrokenMetadata class introduced. Checks script_id and scheduler_id for correct format before displaying. A consistent "BROKEN" state is also displayed for all invalid jobs.

Stopped JobDecorator delegating so many methods to Metadata/BrokenMetadata to stop warnings appearing when running spec, as they are overwritten by methods defined in JobDecorator

Loading jobs using FlightJob::Job.load_all now also loads invalid jobs. FlightJob::Script#serializable_hash was also using load_all so a filter to select only valid jobs was added to that method. Reasoning: The webapp isn't ready to display invalid jobs, and some thought may need to be given to how it will display them, if it ever does.


lib/flight_job/decorators/job_decorator.rb
```diff
@@ -82,15 +88,16 @@ def initialize(object)
...
+        :stdout_path, :stderr_path, :reason, :submit_status, :submit_stdout,
+        :submit_stderr
 
       def created_at(stringify: false)
         time = object.metadata['created_at']
         time && !stringify ? Time.parse(time) : time
       end
 
       def state
+        return 'BROKEN' unless object.valid?
```
benarmston, line 100: This call to `object.valid?` should be cached too.

lib/flight_job/models/job.rb
```diff
@@ -173,7 +178,7 @@ def state
       when 'FAILED_SUBMISSION'
         'FAILED'
       else
-        metadata.state || 'UNKNOWN'
+        valid? ? metadata.state || 'UNKNOWN' : broken_metadata.state
```
benarmston, line 181: Did you come across any issues requiring this line to be added?  If we remove the call to `broken_metadata` from `Job#state` then the only class using `BrokenMetadata` is `JobDecorator`.  `BrokenMetadata` at that point is used only during the display of jobs.

We could also replace `UNKNOWN` with `BROKEN`.  The job's state being unknown sounds quite broken to me.

lib/flight_job/models/job.rb
```diff
@@ -182,7 +187,7 @@ def state
       # This error can be ignored, as the metadata is validated independently
       Flight.logger.error "Failed to resolve the state for job '#{id}'"
       Flight.logger.debug $!
-      return 'UNKNOWN'
+      valid? ? metadata.state || 'UNKNOWN' : broken_metadata.state
```
benarmston, line 190: Similar to above just here please.

lib/flight_job/models/job/broken_metadata.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      def state
+        "BROKEN"
+      end
+
+      %w(script_id scheduler_id).each do |att|
+            define_method(att) do
+              value = @parent.send(att)
+              return value if value.is_a?(Integer) || value.is_a?(String)
+            end
```
benarmston, line 46: The indentation seems odd here.

lib/flight_job/models/job.rb
megan-godfrey: I changed this line to keep the behaviour of the code the same as much as possible, except for introducing the broken state for invalid jobs. So there wasn't a particular situation this was built around - I've changed it

benarmston:
Thanks for making those changes @megan-godfrey.  This LGTM now.

PR #102, opened 2022-05-26
Dev/filter list jobs

megan-godfrey:


benarmston (changes requested):
This is working fairly well.  The UX has a couple of areas that could do with some improvements.

### There are some matching issues

Jobs which are broken due to having an invalid state are not matched by the `--state broken` filter.  Using the decorated state, see `Decorators::JobDecorator`, might fix this.

Array jobs are not correctly matched by the `--state` filter.  This is due to their state being derived from the state of their tasks.  See `Decorators::JobDecorator` for how this is done.

The filters `--state='completed,failed'`, `--state='completed, failed'` and `--state='completed , failed'` behave differently.  They should all work identically to `--state='completed,failed'`.  The same applies to `--id` and `--script` filters.

### The help output also isn't quite right

A command summary should contain only a single line, otherwise the output of `help` looks awkward.  The additional text you added is good, but it should be become the command's `description`.  It will then appear in the output of `help list-jobs`.

The `help list-jobs` output doesn't really indicate that multiple globs can be given by separating with commas.  Adding a `meta: 'ID[,ID...]'` (or something similar) would help.  Or expanding on the option description, e.g., `Filter by job ID. Multiple values can be given by...`.  Or both.

The options are defined as `c.slop.string ...`, they could have alternatively been defined as `c.slop.array ..., delimeter: ','`.  Did you consider that alternative?  Does it provide better help text or not?

Including the list of valid values for the `--state` option would be good.

megan-godfrey:
I experimented with using `c.slop.array`, but I found that it introduced problems without making improvements
-The help wasn't any more helpful
-It also made things more awkward when inputting filters surrounded by quotes - two filters separated by a comma were read in as one element of the array if they were surrounded by quotes, or two elements if they weren't.

lib/flight_job/models/job.rb
```diff
@@ -102,6 +107,20 @@ def save
...
+        if opts[key]
+          return false unless super(opts[key],job_attributes[key])
+        end
+      end
+      true
+    end
+
+    def job_attributes
+      @job_attributes ||= OpenStruct.new(id: id, script: script_id, state: self.decorate.state)
+    end
```
benarmston, line 122: I'm not sure that having `Matcher` defined as a module is the best way to do this.  The use of `super` halfway through `Job#pass_filter?` was a little confusing and its not clear what `Job#job_attributes` is for.

I would have expected `Job#pass_filter?` to be something more like


```ruby
def pass_filter?(opts)
  @_matcher ||= Matcher.new(self.decorate, opts, {id: :id, script: :script_id, state: :state})
  @_matcher.matches?
end
```

or perhaps

```ruby
def pass_filter?(filters)
  @_matcher ||= Matcher.new(filters, {id: id, script: script_id, state: decorate.state})
  @_matcher.pass_filter?
end
```

This much shorter `pass_filter?` method has `Job` declare what is to be filtered and how the option names match to the job values.  That way when we extend filtering to tasks, templates and/or scripts, very little code is required in `Task`, etc. as most all of the implementation is in `Matcher`.

Does that make sense?

lib/flight_job/matcher.rb
```diff
@@ -0,0 +1,48 @@
...
+      false
+    end
+
+    private
+
+    def standardize_string(str)
+      str ||= ""             # Replace nil values with empty string
+      str.downcase           # Case insensitive matching
+         .strip              # Trims white space
+         .gsub(/[\s_]/, '-') # Treat underscores as hyphens
```
benarmston, line 45: Do these comments add anything that the code itself doesn't?  If you think they do, that's fine, but in that case can you update  the `gsub` comment to mention whitespace too.

benarmston (changes requested):
The behaviour and help text looks good to me now.  Thanks for making those changes.  I have a few more comments for you to look at please.

(I pushed a commit to this branch, please pull before you push).

lib/flight_job/matcher.rb
```diff
@@ -0,0 +1,65 @@
...
+# along with FlightHowto. If not, see:
+#
+#  https://opensource.org/licenses/EPL-2.0
+#
+# For more information on FlightHowto, please visit:
+# https://github.com/openflighthpc/flight-howto
+#==============================================================================
+module FlightJob
+  class Matcher
+    attr_reader :filters, :attrs
```
benarmston, line 29: You've ran into one of my pet peeves with Ruby here: overuse of `attr_reader` (and more generally `attr_accessor`).  I'll admit that it's very common to create reader (and accessor) methods for instance variables, but I don't like it.  I don't like it because code is easier to refactor the smaller its public interface is.

For instance, when the interface for `Matcher` is simply `Matcher::new` and `Matcher#matches?` it is (1) very clear how that class can be used; and (2) as long as those two methods continue to work unchanged, we can make as many changes to its internals as we wish knowing that none of the `Matcher` clients will need to be updated.

If internals such as `@filters` and `@attrs` are made public through `attr_reader` then even something as simple as renaming `attrs` has the potential to break clients.

Please either use instance variables directly or make the readers private by moving them below the `private` keyword.

benarmston (changes requested):
Thanks for making those changes @megan-godfrey .  I've got another picky change for you to make please, then this is good to go.

benarmston (approved):
LGTM. Thanks @megan-godfrey

PR #105, opened 2022-06-09
Enh/refactor script metadata

megan-godfrey:


lib/flight_job/models/script.rb
```diff
@@ -114,19 +108,15 @@ def exists?
     end
 
     def notes
-      @notes ||= if File.exist? notes_path
-                   File.read notes_path
-                 else
-                   ''
-                 end
+      @notes ||= Notes.new(id).tap { |n| n.read }
```
benarmston, line 111: I'm not keen on automatically reading the notes here.  Firstly `Notes.new(id)` just looks much nicer than `Notes.new(id).tap { |n| n.read }`; secondly reading the notes from disk into memory is potentially a memory hungry and time consuming process.  Especially when dealing with multiple scripts.

I'm not sure that we needlessly call the `notes` method at the moment, but given reading the notes requires `notes.read` having them automatically read seems awkward.

IIUC, they are read here to ensure that they are set to a string.  If they aren't saving the notes to disk fails.  I think that this can be better handled by doing the following in `Notes#save`:

```ruby
File.write path, @notes || ''
```

lib/flight_job/models/script.rb
```diff
@@ -114,19 +108,15 @@ def exists?
...
     end
 
-    def script_path
-      @script_path ||= File.join(FlightJob.config.scripts_dir, id, 'script.sh')
+    def initialize_notes(notes = "")
+      @notes ||= Notes.new(id).tap { |n| n.write(notes) }
```
benarmston, line 115: I'm not keen on needing an explicit call to `write` here.  I think `Notes.new(id, notes)` would be better.  Where the `notes` parameter is optional.

lib/flight_job/models/script/metadata.rb
```diff
@@ -59,15 +60,10 @@ class Metadata < Metadata::BaseMetadata
...
-          path_tag = File.exist?(@path) ? @path : @parent.id
-          FlightJob.logger.info("Invalid metadata: #{path_tag}\n")
-          JSONSchemaErrorLogger.new(schema_errors, :info).log
-          errors.add(:metadata, 'is not valid')
-        end
-      end
+      validates_with JsonSchemaValidator,
+                     schema: SCHEMA,
+                     json_method: :to_hash,
+                     error_key: :metadata
```
benarmston, line 66: This looks good, but its highlighted that we've lost validation of a scripts metadata.  Can you add the following to `Script` please.

```ruby
    validate do
      unless metadata.valid?
        messages = metadata.errors.map { |e| e.message }
        errors.add(:metadata, messages.join("; "))
      end
    end
```

benarmston: It should then be possible to remove the explicit calls to `script.metadata.valid?` in the specs as those validations should run on the call to `script.valid?(:load)`.

benarmston (changes requested):
I've added a few comments for you to look at.  Mostly this is looking good and the script class is looking much improved from where it was 5 weeks ago :-)

---

Whilst reviewing this I spotted another refactoring that should be done.

* Change `Script::load_all` so that it no longer creates a script object with id `*`.  It should do something similar to `Job::load_all`.

lib/flight_job/models/script.rb
```diff
@@ -213,6 +211,23 @@ def metadata_path
...
+    def initialize_notes(notes)
+      @notes = Notes.new(id, notes)
+    end
+
+    def notes
+      @notes ||= if File.exist?(notes_path)
+                   Notes.new(id, Notes.load_from_path(notes_path))
+                 else
+                   Notes.new(id)
+                 end
```
benarmston, line 223: Class methods called `load*` ought to be returning an instance of the class, also the branching logic here seems like it would better live in a class method on `Note`.  I think this would read much better if it was

```ruby
def notes
  @notes ||= Notes.load(id)
end
```

`Notes::load` would then be responsible for constructing the path, instead of both `Note` and `Script` currently doing so.

lib/flight_job/models/notes.rb
```diff
@@ -25,26 +25,29 @@
...
 module FlightJob
-  class Notes
-    def initialize(script_id)
-      @script_id = script_id
-    end
+  class Script < ApplicationModel
+    class Notes
+      def self.load_from_path(path)
+        File.exist?(path) ? File.read(path) : ''
+      end
```
benarmston, line 32: See comment on `Script#notes`.

lib/flight_job/models/notes.rb
```diff
@@ -25,26 +25,29 @@
...
 
-    def save
-      File.write path, @notes
-      FileUtils.chmod(0600, path)
-    end
+      def save(notes = nil)
+        File.write path, notes || @notes || ''
+        FileUtils.chmod(0600, path)
+      end
```
benarmston, line 46: `Note#{read,save}` should have the following invariant:

```ruby
notes = Notes.new(...)
notes.write(...) == notes.read
```

That is, if we write some notes, calling `read` should return what was just written.

benarmston (changes requested):
I'm being picky again I'm afraid.  Could you look at the following comments please.

benarmston (changes requested):
I've also just noticed that creating scripts is now broken: the validation of the metadata should only happen for persisted jobs. Could you make that change please.

benarmston:
@megan-godfrey I've pushed a commit that I think deals with the `Note::load` / `Note#load` issue.  The commit delays reading notes until the `read` method is called, which avoids any branching at note creating time.  Take a look and let me know what you think.

megan-godfrey:
That seems much better/simpler :) makes sense to only load the notes in when they need to be read, rather than having the two separate load and read methods

benarmston (approved):
LGTM

